PR #19, opened 2022-07-09
[progress bar] How to use section

schmit:
Added a how to use section on the progress bar, as many folks are confused on how to use it. Hopefully this is helpful. Curious to get suggestions on how to make things clearer.


docs/interpreting-experiments/experiment-progress-bar.md
```diff
@@ -1,12 +1,14 @@
 # Experiment Progress Bar
 
+The progress bar helps you understand how much data you have gathered and whether it makes sense to continue gathering more data.
+
 #### Minimum Detectable Effect
-The progress of an experiment is measured with respect to the minimum detectable effect (MDE). This is the minimum effect the experimenter is interested in being able to detect and can be set at the metric level.
+The progress of an experiment is measured with respect to the minimum detectable effect (MDE). This is the minimum effect the experimenter is interested in being able to detect and is set at the metric level. However, it is possible to override this and set it to a particular value at the experiment level.
```
bkfunk, line 6: Is "This is the minimum effect SIZE" more clear?

docs/interpreting-experiments/experiment-progress-bar.md
```diff
@@ -1,12 +1,14 @@
...
 
 **Note:** A small MDE requires the experiment to run much longer to collect sufficient data.
 
 ### Progress Bar
-Each experiment has an associated progress bar at the top. The goal of the progress bar is to measure whether we have gathered sufficient data to detect a lift of the primary metric by the MDE with high probability.
+Each experiment has an associated progress bar at the top. The goal of the progress bar is to measure whether we have gathered sufficient data to detect a lift of the primary metric by a configurable MDE with high probability.
```
bkfunk, line 11: I think the second sentence is a bit unclear. What about something like "The goal of the progress bar is to measure whether we have gathered enough data to make it highly likely that we can detect a meaningful lift in the primary metric, where _meaningful_ means a lift of at least the MDE value configured for that experiment and that metric."

docs/interpreting-experiments/experiment-progress-bar.md
```diff
@@ -18,7 +20,29 @@ You can also see the progress bar on the details page for each experiment.
 
 **Note:** We compute the days remaining using a linear interpolation. This interpolation does not take into account that gathering data usually slows down during an experiment, and so the estimate may be optimistic, especially in the early days of an experiment.
 
-#### How we calculate the progress
+### How to use the progress bar
+
+Traditionally, when using a fixed sample test, you decide up front how long the experiment ought to run and cannot interpret results until you have finished gathering all data. However, the sequential testing approach we use allows for flexibility in deciding when to stop an experiment. Here's some advice on how to get the most out of the progress bar.
+
+Before starting with any experiment, we have to set the minimum detectable effect to something that makes sense for your use case. Two aspects are important:
```
bkfunk, line 27: Flipping between "we" and "you" here when I think both should have the same antecedent? So "Before starting the experiment, you should set the minimum detectable effect to something that makes sense for your use case." Also, if we spelled out and then abbreviated "MDE" above, why not use the abbreviation here?

docs/interpreting-experiments/experiment-progress-bar.md
```diff
@@ -18,7 +20,29 @@ You can also see the progress bar on the details page for each experiment.
...
+- What constitutes a meaningful impact to the business: a 0.1% increase in revenue can be valuable for a company that operates at the scale of Google, but for a start-up that looks to grow 3x year over year, this would make for a negligible impact. 
+- What effect sizes are you able to detect in a reasonable amount of time given the data volume. It's important to keep in mind that, as a rule of thumb, detecting an effect twice as small takes 4 times as long. Thus if based on your data volume it takes a month to detect a 10% lift, then it would take 4 months to detect a 5% lift. A constraint on the maximum time you are comfortable running an experiment often limits what MDE you can aim for.
+
+Keeping the above points in mind, we have set the MDE to a reasonable value, for illustrative purposes let's say at 5%, and we started our experiment. Now how do we use the progress bar? In two ways:
```
bkfunk, line 31: Again, I would try to remain consistent in pronouns.

"Suppose, given the above considerations, you decide to set your MDE at 5%. You start your experiment, and the progress bar begins to fill up. How should you use it to make decisions?

1. The _color_ of the progress bar indicates whether or not the _primary_ metric is significant and positive. Blue indicates it is, gray indicates it is not.

1. The progress bar is an indicator of _when it is safe to conclude there was no effect_. If it is reaching 100% progress, and still gray|grey, it is likely that there is no lift in your primary metric of at least the configured MDE size. Other metrics might show significant results, and might warrant shipping the treatment even though the primary metric is flat, however.
2. If the progress bar is _blue_, it means that your primary metric is better in the treatment variant, and that the difference is statistically significant. Because of the sequential analysis methods we use, _it is safe to stop the experiment and ship it, even if the progress bar is not close to full_.

docs/interpreting-experiments/experiment-progress-bar.md
```diff
@@ -18,7 +20,29 @@ You can also see the progress bar on the details page for each experiment.
...
+
+2. Whenever you see the progress bar turn blue, or see significant effects in your experiment results (the former only consider primary metrics and takes more time) then you can reasonably make a decision based on observed data, even if the progress bar is not filled up at all. For example, if you see a lift of 20%, we know that we can detect such big lift about 4x sooner than the 5% lift we set as minimum detectable effect; we would see a positive result even though the progress bar is only at 25%. 
+![Example when to stop with positive result](../../static/img/measuring-experiments/progress-bar-positive-result.png)
+In this example, we see a roughly 13% increase in the primary metric, and we have gathered enough data to be powered to detect an 11% effect; even though we only see 20% progress towards detecting a 5% lift, we can stop this experiment early and conclude that the new variant improves upon the control.
+
+In short: do not be afraid to conclude an experiment (long) before reaching 100% progress when you see significant results.
```
bkfunk, line 41: I think this section is a bit confusing because different %ages are being thrown around a lot, and we're using example %ages that don't match the images.

bkfunk:
I think this section might be improved if we were just more explicit about the actual logic behind the progress bar, and then translated that logic to decision making recommendations. So, like, literally, what makes it go blue? What makes it go up?

docs/interpreting-experiments/experiment-progress-bar.md
schmit: I like it!

docs/interpreting-experiments/experiment-progress-bar.md
schmit: Good call out; I'm going to stick with we/our

schmit:
Thanks for the feedback @bkfunk, awesome stuff! I made a bunch of edits. Re your last point around being more explicit about the coloring logic, I think that makes sense though it sounds like you might have something specific in mind; want to take a crack at it?

docs/interpreting-experiments/experiment-progress-bar.md
schmit: I agree with your sentiment, but not quite sure how to fix it at the moment; I imagine the examples are useful over not having anything at all; should we ship as is and then improve further in a next iteration?

PR #21, opened 2022-07-11
redshift ssh tunnel

leoromanovsky:


docs/connecting-your-data/data-warehouses/connecting-to-redshift.md
```diff
@@ -72,6 +72,23 @@ You'll want to gather the following connection details from Redshift:
...
+
+* tunnel host
+* username & password
+
+**SSH tunnel with public key**
+
+* tunnel host
+* username
+* public key
+
```
chasdevs, line 91: Should we add a screenshot or some actual example? It always makes me feel better when I'm reading docs and there's a specific example.

chasdevs (approved):

docs/connecting-your-data/data-warehouses/connecting-to-redshift.md
```diff
@@ -1,6 +1,6 @@
 # Connecting to Redshift
 
-## 1. Whitelist Eppo IP Address
+## 1. Allowlist Eppo IP Address
```
schmit, line 3: Nit: I like this change but allowlist sounds more like a noun than a verb to me, making this sound a little odd, maybe just `Allow Eppo IP address`?

schmit (approved):
As a famous Leo likes to say: Good stuff!

docs/connecting-your-data/data-warehouses/connecting-to-redshift.md
petzel: Could also be `Add Eppo IP addresses to Allowlist`

PR #26, opened 2022-07-14
Improvements to metrics docs

dylandaniels:
* Descriptions of each type of aggregation
* Removed reference to metric-based filters (no longer exist)
* Minor rewordings

### DIFF
--- a/docs/building-experiments/metrics/creating-metrics.md
+++ b/docs/building-experiments/metrics/creating-metrics.md
@@ -42,7 +42,7 @@ The minimum detectable effect refers to the smallest effect you want to reliably
 
 ### Sum
 
-Sum computes metrics that are typically interpreted as averages per entity. If the fact value is NULL, it is imputed as 0.
+Sum computes metrics that are typically interpreted as averages per entity. If the fact value is NULL, it is discarded.
 
 $\frac{\text{SUM of fact value}}{\text{Number of unique entities}}$
 
@@ -56,6 +56,7 @@ $\frac{\text{Sum of (COUNT DISTINCT of fact value for each unique entity)}}{\tex
 
 Examples: unique videos watched per user, unique articles viewed per visitor, unique products viewed per user.
 
+
 ### Count
 
 Count leverages SQL's ``COUNT`` to compute a total count of events per entity. If the fact value is NULL, it is discarded.
@@ -68,14 +69,22 @@ Examples: videos watched per user, articles viewed per visitor, orders per user.
 
 Retention metrics measure the proportion of entities who have at least one fact value appear after a fixed number of days (X) from experiment assignment. For example, a 7-day retention metric might measure the proportion of users who re-visit a website at least 7 days after being assigned to the experiment. 
 
-$\frac{\text{Sum of \{1 if fact value is non-null after X days of assignment, else 0, for each unique entity\}}}{\text{Number of unique entities}}$
+$\frac{\text{Sum of \{1 if a non-null fact value is present X days after the assignment time, else 0, for each unique entity\}}}{\text{Number of unique entities}}$
+
+For example, if $X = 7 \text{ days}$, Eppo records a retention event for an entity when
+
+$(\text{timestamp of event}) - (\text{timestamp of assignment}) >= 7 \text{ days}$. 
 
 
 ### Conversion
 
 Conversion metrics measure the proportion of entities who have at least one fact value appear within a fixed number of days (X) from experiment assignment. For example, a 7-day conversion metric might measure the proportion of users who sign up for a free trial within 7 days of being assigned to the experiment. 
 
-$\frac{\text{Sum of \{1 if fact value is non-null within X days of assignment, else 0, for each unique entity\}}}{\text{Number of unique entities}}$
+$\frac{\text{Sum of \{1 if fact value is non-null within X days of the assignment time, else 0, for each unique entity\}}}{\text{Number of unique entities}}$
+
+For example, if $X = 7 \text{ days}$, Eppo records a conversion event for an entity when 
+
+$(\text{timestamp of event}) - (\text{timestamp of assignment}) < 7 \text{ days}$. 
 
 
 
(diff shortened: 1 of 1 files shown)


dylandaniels:
Would like review on:

(1) Whether current descriptions are accurate
(2) Behavior of conversion/retention metrics before the specified number of days has passed.
(3) Treatment of NULL fact values.

docs/building-experiments/metrics/creating-metrics.md
```diff
@@ -34,14 +34,48 @@ You may actually want to create a metric that is a ratio of two facts. To do thi
...
+
+Examples: average revenue per user, sign-up rate, minutes streamed per user, average order value.
+
+### Count Distinct
+
+Count Distinct leverages SQL's ``COUNT DISTINCT`` to compute a unique average per entity. If the fact value is NULL, it is discarded.
+
+$\frac{\text{Sum of (COUNT DISTINCT of fact value for each unique entity)}}{\text{Number of unique entities}}$
+
+Examples: unique videos watched per user, unique articles viewed per visitor, unique products viewed per user.
```
schmit, line 57: It depends a bit on the fact, but I think this is incorrect, usually using count distinct gives you "Number of users who have watched (at least) a video"

docs/building-experiments/metrics/creating-metrics.md
```diff
@@ -34,14 +34,48 @@ You may actually want to create a metric that is a ratio of two facts. To do thi
...
+
+Count leverages SQL's ``COUNT`` to compute a total count of events per entity. If the fact value is NULL, it is discarded.
+
+$\frac{\text{Sum of (COUNT of fact value for each unique entity)}}{\text{Number of unique entities}}$
+
+Examples: videos watched per user, articles viewed per visitor, orders per user.
+
+### Retention
+
+Retention metrics measure the proportion of entities who have at least one fact value appear after a fixed number of days (X) from experiment assignment. For example, a 7-day retention metric might measure the proportion of users who re-visit a website at least 7 days after being assigned to the experiment.
```
schmit, line 69: Maybe add a bit of detail: 
```suggestion
Retention metrics measure the proportion of entities who have at least one fact value appear after a fixed number of days (X) from experiment assignment. For example, a 7-day retention metric on the website visits fact might measure the proportion of users who visit a website at least 7 days after being assigned to the experiment. 
```

docs/building-experiments/metrics/creating-metrics.md
```diff
@@ -16,15 +16,15 @@ This should be one of the facts that you created in the step above, and should c
 
 ![Select Aggregation](../../../static/img/building-experiments/select-aggregation-for-metric.png)
 
-The aggregation will aggregate over whatever the fact is measuring on a per-entity basis. So for example, if you select the `Revenue` fact and the `SUM` aggregation, the metric will be the total revenue for each user;, if you select the `Upgrades` fact and the `SUM` aggregation, the metric will be the total number of upgrades for each user, which should only be 1 across the board.
+The aggregation will aggregate over whatever the fact is measuring on a per-entity basis. So for example, if you select a `Revenue` fact and the `SUM` aggregation, the metric will be "total revenue per user;" if you select a `Name of Article Viewed` fact and the `COUNT DISTINCT` aggregation, the metric will be "Unique articles viewed per user."
```
schmit, line 19: Should we be clear that we're always taking the average across entities, e.g. "average total revenue per user"

schmit (approved):
Awesome additions! left a few nitty comments

docs/building-experiments/metrics/creating-metrics.md
```diff
@@ -34,14 +34,48 @@ You may actually want to create a metric that is a ratio of two facts. To do thi
...
+Retention metrics measure the proportion of entities who have at least one fact value appear after a fixed number of days (X) from experiment assignment. For example, a 7-day retention metric might measure the proportion of users who re-visit a website at least 7 days after being assigned to the experiment. 
+
+$\frac{\text{Sum of \{1 if fact value is non-null after X days of assignment, else 0, for each unique entity\}}}{\text{Number of unique entities}}$
+
+
+### Conversion
+
+Conversion metrics measure the proportion of entities who have at least one fact value appear within a fixed number of days (X) from experiment assignment. For example, a 7-day conversion metric might measure the proportion of users who sign up for a free trial within 7 days of being assigned to the experiment. 
+
+$\frac{\text{Sum of \{1 if fact value is non-null within X days of assignment, else 0, for each unique entity\}}}{\text{Number of unique entities}}$
```
schmit, line 78:
```suggestion
$\frac{\text{Sum of \{1 if fact value is non-null within X days of assignment, else 0, for each unique entity\}}}{\text{Number of assignments}}$
```

Maybe assignments is better than entities here? (entity is "User" vs "Visitor"). But this is correct. For both retention and conversion we do count all assignment, even if they haven't been in assigned for the entire period (e.g. assigned 3 days ago with a 7 day window)

docs/building-experiments/metrics/creating-metrics.md
dylandaniels: @schmit I thought that "Number of users who have watched (at least) a video" would require a threshold metric? If I have a fact that lists a unique string associated with each video, and I perform a count distinct over them, couldn't I get numbers higher than 1 for each user?

docs/building-experiments/metrics/creating-metrics.md
dylandaniels: Going to change this to "unique entities assigned" to be super super clear

docs/building-experiments/metrics/creating-metrics.md
dylandaniels: Making this change

docs/building-experiments/metrics/creating-metrics.md
schmit: That's a great call-out; your thinking makes a lot of sense but the code disagrees with you:
https://github.com/Eppo-exp/eppo/blob/main/packages/api/src/console/db-aggregations/experiment-subject-summaries.blocks.ts#L76

dylandaniels: Wow, proven wrong! I'm going to say this is an objectively bad name for this function but document this correctly. It's confusing that:
* `COUNT` is a count of the number of events per user
* `COUNT DISTINCT` is the number of users with an event (rather than distinct events per user)

PR #35, opened 2022-07-25
Update alpha value for traffic imbalance

schmit:
We currently use an alpha = 0.01 instead of 0.001


evanmiller:
Fixed in https://github.com/Eppo-exp/eppo/pull/4099

PR #44, opened 2022-08-05
Delete FAQ file

dylandaniels:
Empty file which does not answer the question it poses.


schmit (approved):
Oh yes thanks! This was bugging me too

PR #49, opened 2022-08-22
Added a relationship map

Summchestnuts:
@megan-eppo Please let me know if this looks alright. Not sure how to resize the image, if possible, we should have the map a bit small but it can be clicked to view the full image.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -10,6 +10,8 @@ Follow this quickstart to get your first experiment set up on Eppo.
 
 ## Overview
 
+![Eppo Product Relationship Map](https://user-images.githubusercontent.com/111296875/186039257-d2d9f8db-e42a-45e7-af5f-19c899d54339.jpg)
+
 0. [Prep your data warehouse and set up feature flagging](#0-prep-your-data-warehouse-and-set-up-feature-flagging)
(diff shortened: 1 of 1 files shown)


megan-eppo (commented):
You can try seeing if wrapping the image in html works
`Eppo Product Relationship Map`

Summchestnuts:
> You can try seeing if wrapping the image in html works `Eppo Product Relationship Map`

I tried, and it didn't work, but maybe I'm not getting this right... Let me ask about around and figure it out

vpai:
@Summchestnuts this is a great diagram! But I feel like if I'm new to the Eppo and reading this I'm very confused by all the complexity. One small step here that might help a bit would be adding the numbers of the list below to the actual diagram, that way I can map the diagram to the text below easily. Otherwise I feel like we're leaving people hanging by introducing a really complex picture to an already complex page.

Summchestnuts:
That is an awesome idea!!!!!!! I’ll edit the map and update this. Thank you!

Summchestnuts:
@vpai what do you think about this?
Summer's random board - Relationship Map

Summchestnuts:
@vpai do you want to change the map to the one above with numbers matching the overview?

PR #61, opened 2022-09-22
Add documentation for Bayesian CI and revise lift estimate and confidence interval docs in general

bkfunk:
# Goal of this PR

After comments on the initial draft, I have revised (and extended) this PR to edit, rearrange, and add documentation for:

- Setting and understanding analysis plans, with particular attention to providing guidance on how to choose analysis method
- Using the experiment list and details pages
- Using the statistical details view, including documentation for the different statistics presented under both frequentist and Bayesian methods
- Understanding the basics of lift estimates and CIs
- In-depth documentation of each analysis method

The goal is to, primarily, add documentation for the Bayesian analysis method, flesh out technical documentation for analysis methods and analysis plans that is already there in basic form, and provide a description of our methods at varying levels of granularity and technical expertise, so that we can have pages that work for PMs, pages that work for non-statsy eng, and pages that work for DS.

# Overview of content changes

Most of the new content is in `docs/experiments/interpreting-experiments`:

- Added a lot of content to [`docs/experiments/interpreting-experiments/index.md`](https://github.com/Eppo-exp/eppo-docs/pull/61/files?diff=split&w=0#diff-c87f880a3d13e6c17023ddb0f21ba431f76283d197f8ac2300b142bf6eec3550) to provide an overview of the UI for the main experiment list and details pages. This page is targeted toward non-technical users of Eppo who are trying to just use the product, and the goal is to understand not just how to navigate to different surfaces but the basics of how to use the information on those surfaces. In addition, I updated most of the screenshots on this page to reflect the current UI and also show surfaces that were not previously shown.
- Broke the content in [`docs/experiments/interpreting-experiments/confidence-intervals.md`](https://github.com/Eppo-exp/eppo-docs/pull/61/files?diff=unified&w=0#diff-478c67c05de44e23bf2339bf022bac148968b7155a55e485dcf1d5944d4059e8) into multiple pieces with in the folder `docs/experiments/interpreting-experiments/lift-estimates-and-confidence-intervals/`
  - The main [`docs/experiments/interpreting-experiments/lift-estimates-and-confidence-intervals/index.md`](https://github.com/Eppo-exp/eppo-docs/pull/61/files?diff=unified&w=0#diff-21773778d619e0ac013eb9fd84e001751fbaacd207abf60cdb09d0829f7cf474) page provides a broad overview of what "lift" and "confidence intervals" mean, without too many technical details
  - Added [`docs/experiments/interpreting-experiments/lift-estimates-and-confidence-intervals/analysis-methods.md`](docs/experiments/interpreting-experiments/lift-estimates-and-confidence-intervals/analysis-methods.md) to cover the differences between the analysis methods and why you might choose one over another. This page has some technical depth, but is designed for a data scientist or possibly PM who simply wants to know how to set up their experiment.
  - Added [`docs/experiments/interpreting-experiments/lift-estimates-and-confidence-intervals/statistical-nitty-gritty.md`](docs/experiments/interpreting-experiments/lift-estimates-and-confidence-intervals/statistical-nitty-gritty.md) to go into all the technical details, step by step. The audience for this page is a DS or other data/eng leader involved in evaluating the POC; they might have familiarity with their current experimentation methods and want to understand the differences, or to replicate our methods locally to assuage concerns about us being a black box, or just get a sense of how sophisticated our methods are and how they compare to competitors. Although most people using this page will have significant stats knowledge, I wanted technical users who maybe took stats in college but haven't used it much in a while to still be able to grok what we're doing (and why). 
- Added a page dedicated to analysis plans [`docs/experiments/planning-experiments/analysis-plans.md`](https://github.com/Eppo-exp/eppo-docs/pull/61/files?diff=unified&w=0#diff-8a03adb4daa8cf4ab998c0a599c9408ffe23b907a0b4413de86bbfee23680adc), which describes each setting and provides guidance on how to choose a value for each setting. In order to do this, it also describes the basics of the different analysis methods and provides a broad framework for developing a decision-making process around experimentation.
- Fixed some typos, style inconsistencies, etc., and updated links to reflect above changes.

# Overview of non-content changes

The big drop in total lines is that I removed package-lock.json to move entirely to yarn. I also:

- Add a bunch of utility components in `src/components/`, which simplify tasks I had to do several times in this PR
  - One thing to note: I added [`src/components/glossary/Term.js`](https://github.com/Eppo-exp/eppo-docs/pull/61/files?diff=unified&w=0#diff-9c382e118f896efce5f8631c080864e9aaa73c8c76c29e8a0701a8dd5054489d) and used it in a few places; right now it just serves as a way to apply consistent styling, but eventually, given that I found myself using somewhat esoteric terminology a lot in many places, I think it would be good to add a glossary and have the terms show tooltips to provide brief definitions, so that you don't have to link all over the place if you forgot what, say, the "confidence level" means, exactly. It is probably overengineered now, but I think not worth ripping out
- I added and updated some packages in [`package.json`](https://github.com/Eppo-exp/eppo-docs/pull/61/files?diff=unified&w=0#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519) to fix some rendering bugs, add features like labeled footnotes (so you don't have to change note numbers if you reorder the text, and also it makes it easier to know which note is for which piece of text), and add linting. Linting is not fully implemented now, but I'm leaving that as a `TODO`.
- Changed [`docusaurus.config.js`](https://github.com/Eppo-exp/eppo-docs/pull/61/files?diff=unified&w=0#diff-a038096cbdea434999e1dce5ab497212f1fe18204dde1a027ce3bdd663261a2a) to dynamically load `katex` so I could use a more recent version (which was required for the equation numbering features I wanted). See https://docusaurus.io/docs/markdown-features/math-equations#upgrading-rehype-katex-beyond-recommended-version

# What is requested of reviewers

- For stats people, review the stats language and make sure everything is accurate
- For others, please look for glaring inconsistencies, typos, things that are confusing, broken links, etc.


docs/interpreting-experiments/cuped.md
```diff
@@ -1,7 +1,6 @@
 # Controlled-experiment Using Pre-Existing Data (CUPED)
 The precision of experiment results (the width of confidence intervals) depends on the variance of the metrics we are measuring. One way to improve the precision of experiments is to gather more data, as the variance goes down as we gather more data; this obviously means that it takes longer to run an experiment. 
 
-owever, there are more direct ways to reduce variance, beyond just waiting for more data to show up.
```
bkfunk, line 4: Delete stray line

docs/interpreting-experiments/confidence-intervals.md
```diff
@@ -1,34 +1,266 @@
...
+
+#### Confidence Intervals
+
+For fixed-sample and Bayesian analysis, constructing a normal distribution that describes the average lift (equation 1-3 in the case of fixed-sample, and the posterior from equations 5-6 in the case of Bayesian) is enough to construct confidence intervals: the lower and upper bounds for significance level $\alpha$ (e.g. 5%) are simply the $\frac{\alpha}{2}^{th}$ and $(1 - \frac{\alpha}{2})^{th}$ percentiles of that distribution.
+
+In order to guarantee always-valid confidence intervals for sequential analysis, however, we need to do something fancier. Specifically, we use the method described in [Howard et al. (2021)](https://arxiv.org/abs/1810.08240) to construct a bound such that the probability that the mean will be outside that bounds at any point is less than or equal to a significance level $\alpha$. In particular, we define the *confidence sequence* (that is, a sequence of confidence intervals) around the mean using equation (14) from the reference (with some changes in notation):
+
+$\hat \mu \pm \frac{\sigma}{t} \sqrt{ (t + \rho) \log \left( \frac{t+\rho}{\rho\alpha^2} \right)}$
```
bkfunk, line 260: @schmit Do you remember where the sigma/t comes from, before the sqrt? I've skimmed the paper again but I'm assuming I'm just missing a notation thing

docs/interpreting-experiments/confidence-intervals.md
```diff
@@ -1,34 +1,266 @@
-# Confidence Intervals
-By navigating to the detail view of any experiment, you can view the effect that any of the treatments have had on each of the selected metrics for that experiment. You can get to the detail view of any experiment by clicking the **Experiments** tab on the left panel and then clicking on the experiment you're interested in from the list. The effect on each metric comes with an accompanying confidence interval. This confidence interval tells you the upper and lower bound of the effect on the metric in addition to the point estimate of the effect. We show confidence intervals for the ratio between treatment and control, rather than their absolute difference.
+# Lift Estimates and Confidence Intervals
```
schmit, line 1: I wonder if it's useful to add a short intro paragraph with TLDR and content in this page as it's pretty long

docs/interpreting-experiments/confidence-intervals.md
```diff
@@ -1,34 +1,266 @@
...
-However, this approach assumes that the sample size is fixed up front, and if this assumption is violated the guarantees for the confidence intervals become invalid.
+## How we calculate the lift estimates and confidence intervals
 
-#### Sequential Analysis
-Eppo uses [sequential analysis](https://arxiv.org/abs/1810.08240) which allows us to compute confidence intervals that are valid across time. More precisely, the probability that the random confidence interval does not contain the unknown mean at any point in time is bounded by $\alpha$. We use the bound in equation (14) of the [reference](https://arxiv.org/abs/1810.08240), and generate confidence intervals by computing
+Eppo provides three methods for estimating the lift and calculating confidence intervals:
 
-$\hat \mu \pm \frac{\sigma}{t} \sqrt{ (t + \rho) \log \left( \frac{t+\rho}{\rho\alpha^2} \right)}$
+1. Fixed-sample frequentist
+2. Sequential frequentist (*the default*)
```
schmit, line 22: Why not put the default as number 1?

docs/interpreting-experiments/confidence-intervals.md
```diff
@@ -1,34 +1,266 @@
...
 
-If the change is positive and a positive change has been marked as good, the interval shows up in green (red if positive is bad). The opposite is true if the change is negative. The interval will be gray if the confidence range includes 0, possibly indicating that the effect of the variation on that metric was not significant.
+The effect on each metric comes with an accompanying **confidence interval**, showing the range of values we consider plausible at the specified **significance level**, which defaults to 5%. The significance level represents how likely a treatment that was identical to the control (that is, an A/A test) would show an effect at least as large as what you've observed in your experiment.[^1] A *higher* significance level means that you are willing to accept more risk of seeing an effect that isn't actually there[^2], but also that you can make a decision with less data; it will also mean that the confidence intervals will be narrower.
```
schmit, line 4: I think we should avoid this language around significance level; the whole point of using confidence intervals is to hide this statistical jargon from folks. 

Instead, we should just rely directly on the definition of a confidence interval: we generate an interval that contains the true underlying value 95% of the time, with slight differences depending on the method chosen (this is not really true for the Bayesian method)

schmit: I think it's a useful exercise to rewrite this and avoid the term "significance level" and instaed use "confidence level" (95% instead of 5%)

docs/interpreting-experiments/confidence-intervals.md
```diff
@@ -1,34 +1,266 @@
...
+
+If zero falls within the confidence interval, then we cannot rule out (at the given significance level) the possibility that the treatment is no different from control. In that case, the confidence interval will be colored gray. On the other hand, if the entire confidence interval is above or below zero, then the effect is *statistically significant*. In that case, the confidence interval and the lift estimate box will be colored <span class="positive-change-green-bg">green</span> if the change is good, and <span class="negative-change-red-bg">red</span> if the change is bad. If the change is good, there will also be a 🎉 symbol in the lift estimate box.
```
schmit, line 6: Here too I think we can try to avoid statistically significant -- we can talk about the connection with hypothesis tests in the more detailed section below.

docs/interpreting-experiments/confidence-intervals.md
```diff
@@ -1,34 +1,266 @@
...
+1. Fixed-sample frequentist
+2. Sequential frequentist (*the default*)
+3. Bayesian
 
-where $\hat \mu$ is the estimate for the relevant parameter we are interesting in estimating, $t$ is the number of observations, $\alpha$ is the significance level, and $\rho$ is set using
+Each method has advantages and tradeoffs, which we'll discuss below, but we chose to make sequential analysis the default because it allows you the most flexibility about how long to run experiments and when to make decisions, while relying on fewer assumptions that might not hold in common experimentation scenarios. Fixed-sample analysis may in some cases provide a better ability to detect effects, but it does so at the cost of significant restrictions on how you can run your experiments and is particularly brittle to any of the pre-experiment assumptions being incorrect. Bayesian analysis may allow for a more intuitive way to make decisions, since it does not rely on "statistical significance", but requires making pre-experiment assumptions about what the lift is likely to be, which is hard to do in a consistent way for all experiments and for all companies.
```
schmit, line 25: > is particularly brittle to any of the pre-experiment assumptions being incorrect

This seems vague; I imagine you are trying to say that you have to specify run time in advance and you need to set the MDE correctly. Maybe we can state that more explicitly?

schmit: Or we could reword entirely and say: if you have a culture of running experiments for a fixed amount of time, e.g. 2 weeks, and you avoid peeking and deciding prematurely, then the fixed sample test gives you more power.

docs/interpreting-experiments/confidence-intervals.md
```diff
@@ -1,34 +1,266 @@
...
+
+The analysis itself consists of measuring the mean and variance of the lift you observed and then using those to conduct a [Null Hypothesis Significance Test](https://en.wikipedia.org/wiki/Statistical_hypothesis_testing#The_testing_process). In our case, the null hypothesis is that the treatment and control variants are identical, that is, that you are performing an A/A test; if it is unlikely[^3], if the null hypothesis *were* true, that we would observe a lift between the two variants that is as large as what was measured during the experiment, then we *reject* the hypothesis that both variants are identical and call the lift *statistically significant*.
```
schmit, line 40: > if it is unlikely[^3], if the null hypothesis *were* true, that we would observe a lift between the two variants that is as large as what was measured during the experiment, then we *reject* the hypothesis that both variants are identical and call the lift *statistically significant*.

The double if is hard to grok here

schmit, line 40: Here again I think it makes more sense to focus on confidence intervals, and just mention the relation to hypothesis testing in passing

docs/interpreting-experiments/confidence-intervals.md
```diff
@@ -1,34 +1,266 @@
...
+- **You cannot decide to run the experiment to collect more sample**, even if you do end up with an underpowered experiment after reaching the predetermined sample size, or if, at that point, you decide you want to understand how the experiment affected different segments or subpopulations; doing so would increase your false positive rate above the specified significance level. Instead, you need to start a *new* experiment, and run it longer before making a decision.[^4]
+- **You cannot make decisions before the sample size is achieved** without invalidating the statistical guarantees.[^5] (aka: No peeking! 🙈) If you see a big metric drop, for example, and decide to shut down the experiment early, there is no guarantee that the chance it was a false positive (that is, that the drop was due to pure chance) is less than the significance level you've set. If you do this repeatedly, needlessly shutting down experiments that actually had neutral or even positive results. Similarly, if you want to ship experiments early when metrics seem to go up a lot, you'll end up shipping a lot of treatments with null or negative results.
```
schmit, line 52: `No peeking` -> `no peeking`?

schmit: Also want to move this in front of the period?

schmit, line 52: > If you do this repeatedly, needlessly shutting down experiments that actually had neutral or even positive results.

Something is missing in that sentence

docs/interpreting-experiments/confidence-intervals.md
```diff
@@ -1,34 +1,266 @@
...
+- **You cannot use only part of the data to make a decision.** If there's a data problem at a certain point in your experiment, so you cannot use the full sample size, you need to restart the experiment from scratch; trying to use the data you collected before the problem would cause the same issues as making a ship/shutdown decision before achieving the predetermined sample size. Similarly, if you find that the treatment effect is observed only for some segments of subjects, you cannot make a decision just on that subpopulation without obviating the significance level you've set.
```
schmit, line 53: I don't really get this point -- if you throw away data without looking at outcomes (e.g. let's throw away all data on Tuesdays, ignoring weekly seasonality), then it should not impact the outcome?

docs/interpreting-experiments/confidence-intervals.md
```diff
@@ -1,34 +1,266 @@
...
+- **You can make any decision at any time, safely.** Go ahead, peek away 👀! If a metric is marked as statistically significant (so the confidence interval is entirely above or below zero), you don't have to worry about it being a false positive just because you haven't reached a predetermined sample size. Also, you can decide to keep running the experiment to collect more sample, and, unlike with fixed-sample analysis, doing so will not invalidate the statistical guarantees.
+- **Any data you collect can be useful, even if something goes wrong.** If a data problem means that you can't use data past a certain date, you can at least use all the data you've collected up until then. This means no need to rerun an experiment just because a pipeline broke or there was a bug in logging.
```
schmit, line 74: Similarly to the point above, I dont understand this point

docs/interpreting-experiments/confidence-intervals.md
```diff
@@ -1,34 +1,266 @@
...
+
+#### Cons of sequential analysis
+
+- **You might have less power.** Compared to a fixed-sample analysis, sequential analysis has less power, meaning it's less likely to detect a real effect. However, this loss in power is mostly concentrated in really small sample sizes, and it can be mitigated by using techniques like [CUPED](cuped.md). Furthermore, fixed-sample analysis only has more power if the power analysis used to set the predetermined sample size is actually correct: if the treatment group has a higher than expected variance for a given metric, for example, you'll end up being less powered than you anticipated—and the only thing you can do in that case is rerun the experiment from scratch. With sequential analysis, if you have less power than you'd like you can always choose to keep running the experiment to collect more sample, without affecting the statistical guarantees
```
schmit, line 78: > Compared to a fixed-sample analysis, sequential analysis has less power

I think it may be useful to be a bit more explicit here: for a fixed amount of time, it has less power

schmit: > However, this loss in power is mostly concentrated in really small sample sizes, and it can be mitigated by using techniques like [CUPED](cuped.md)

The loss in power is true for all sample sizes no?

Also, CUPED helps but then again CUPED helps fixed sample too so in the end it's still worse no?

docs/interpreting-experiments/confidence-intervals.md
```diff
@@ -1,34 +1,266 @@
...
+
+The prior we use is described specifically in the [nitty-gritty](#statistical-nitty-gritty) section below, but in essence we set our pre-experiment belief to be that the lift on any given metric will be, on average, zero, that for 50% of experiments the lift will fall between -21% and +21%, and that for 95% of experiments the lift will fall between -62% and +62%; if your experiments tend to show bigger lifts in either direction, then our Bayesian confidence intervals[^6] might be too narrow (biasing toward showing an effect when there is none).
```
schmit, line 86: Side note: I think these priors are slightly wide for what we've seen in data, but we can adjust later

docs/interpreting-experiments/confidence-intervals.md
```diff
@@ -1,34 +1,266 @@
...
+- **You can make a decision based on a nuanced understanding of the probabilities.** Frequentist methods for experimentation inevitably boil down to *statistically significant* vs. *not statistically significant*. With Bayesian results, you have the whole posterior *distribution*, not just a binary "yes" or "no". This means you can make decisions not only on whether the lift is different from zero or not, but also on how likely the treatment is to be better than the control, or what the expected downside would be *if you are wrong*. When an experiment can move metrics in different ways, and different movements in different metrics can have very different costs to the business, this level of nuance allows you to make the best decision for each particular situation.
```
schmit, line 103: > Frequentist methods for experimentation inevitably boil down to *statistically significant* vs. *not statistically significant*.

Well we do have a CI

docs/interpreting-experiments/confidence-intervals.md
```diff
@@ -1,34 +1,266 @@
...
+
+#### Cons of Bayesian analysis
+
+- **Stakeholders might not be used to the Bayesian way of thinking.** Although some Bayesian concepts are more intuitive than their frequentist counterparts, frequentism is still the more common paradigm people learn for doing statistical inference, so switching from "This metric had a statistically significant increase" to "The treatment is 95% likely to have increased this metric" might require educating stakeholders on how to use Bayesian results.
+- **You have to decide *how* to make decisions.** The output of Bayesian inference is a statistical distribution, rather than a "yes"/"no" answer to the question "Was there a statistically significant effect?" This means that decision-makers need to align on how to translate that distribution into a decision on whether to ship a particular treatment. For example, do you care more about risk, or probability the treatment beats control? What probability that the lift exceeds the MDE is considered "good enough to ship"?
```
schmit, line 110: Since we show a credible interval, which is pretty similar to a confidence interval, these can be used in the same way no?

schmit (commented):
Added comments on the sections before the nitty gritty; hope to look at that tomorrow. Overall awesome stuff, love the callouts :)

docs/interpreting-experiments/confidence-intervals.md
```diff
@@ -1,34 +1,266 @@
...
+And similarly for the treatment:
+
+$$
+\lim_{N_T \to \infty}\bar{X}_T \sim \mathcal{N}(\mu_T; \frac{\sigma^2_T}{N_T})
+$$
+
+We don't observe the true means and variances, but we *can* estimate them, using $m_C$ and $s_C$ and $m_T$ and $s_T$:
+
+$$
+\lim_{N_C \to \infty}\bar{X}_C \sim \mathcal{N}(m_C; \frac{s^2_C}{N_C});
```
schmit, line 146: This should be:

$\hat \mu \to N(\mu, \sigma^2/N)$

docs/interpreting-experiments/confidence-intervals.md
```diff
@@ -1,34 +1,266 @@
...
+
+\lim_{N_T \to \infty}\bar{X}_T \sim \mathcal{N}(m_T; \frac{s^2_T}{N_T})
+$$
+
+We want to calculate the lift:
+
+$$
+\hat{\mathcal{L}} = \frac{\bar{X}_T - \bar{X}_C}{\bar{X}_C} = \frac{\bar{X}_T}{\bar{X}_C} - 1
```
schmit, line 154: Nit: I suggest using $\Delta$ instead of $\mathcal{L}$ since the latter is usually used to denote the likelihood. Delta isn't perfect but a bit more common I think

docs/interpreting-experiments/confidence-intervals.md
```diff
@@ -1,34 +1,266 @@
...
+
+#### Bayesian analysis: updating our prior
+
+In a Bayesian framework, you start with a *prior distribution*, which describes what you believe before running the experiment. Then, you run the experiment and collect data, which you use to *update* your prior: in essence, you combine your pre-experiment beliefs about what the lift would be, with the evidence you've gotten from the experiment, into a *new* set of beliefs, called the *posterior* (because it comes *after* gathering data). The estimated average lift is then just the mean of this posterior distribution.
+
+In our implementation of the Bayesian approach, we start the following prior on the lift:
+
+$$
+\begin{equation}
+P(\mathcal{L}) \sim \mathcal{N}(\mu_{pre}=0; ~\sigma_{pre}^2 = 0.1)
```
schmit, line 190: Nit: I think the notation is a bit weird, as P(.) usually indicates a probability. I don't think we can just say the prior is a normal distribution without definiing P(L)

docs/interpreting-experiments/confidence-intervals.md
```diff
@@ -1,34 +1,266 @@
...
+
+&\text{\it\small{{(multiply top and bottom by $\sigma^2_{prior}\sigma^2_{data}$)}}}
+
+\\[10pt]
+
+\sigma^2_{post} =&~~ \frac{1}{\frac{1}{\sigma^2_{prior}} + \frac{1}{\sigma^2_{data}}}
+\end{align}
+$$
+
+Following [Gelman](http://www.stat.columbia.edu/~gelman/book/BDA3.pdf) (p50) (with a slight tweak), we can rewrite the posterior mean $\mu_{post}$ as:
```
schmit, line 235: Unclear whether this is page of pdf or page of book -- might be better to refer to the equation. Also what is the slight tweak?

docs/interpreting-experiments/confidence-intervals.md
```diff
@@ -1,34 +1,266 @@
...
+- $(\mu_{data} - \mu_{prior})$ is the difference between what we've observed and what we expected to observe before we started the experiment
+
+Then we can restate equation (8) as: the lift we observed in the experiment, *shrunk* toward the lift we *expected* to see (according to the prior) by an amount that is set by how precise our lift estimate is relative to the strength of our prior.
+
+[^8]: And it will get to zero faster if our prior is *weaker*, that is, if it has a higher variance. If we have a *strong* belief, represented by a *low* value of $\sigma^2_{prior}$, then moving this term toward zero requires need *more* data (or, data that doesn't vary much).
+
+#### Confidence Intervals
+
+For fixed-sample and Bayesian analysis, constructing a normal distribution that describes the average lift (equation 1-3 in the case of fixed-sample, and the posterior from equations 5-6 in the case of Bayesian) is enough to construct confidence intervals: the lower and upper bounds for significance level $\alpha$ (e.g. 5%) are simply the $\frac{\alpha}{2}^{th}$ and $(1 - \frac{\alpha}{2})^{th}$ percentiles of that distribution.
```
schmit, line 256:
```suggestion
For fixed-sample and Bayesian analysis, constructing a normal distribution that describes the average lift (equation 1-3 in the case of fixed-sample, and the posterior from equations 5-6 in the case of Bayesian) is enough to construct confidence/credible intervals: the lower and upper bounds for significance level $\alpha$ (e.g. 5%) are simply the $\frac{\alpha}{2}^{th}$ and $(1 - \frac{\alpha}{2})^{th}$ percentiles of that distribution.
```

docs/interpreting-experiments/confidence-intervals.md
schmit: IIRC in the paper they analyze a sum of $N(\mu,1)$ distributed random variables, but for us it's $N(\mu, \sigma^2)$ and then we are looking at the variance of the mean, not the sum so that's why we divide by $t$.

schmit (changes requested):
Added feedback on the nitty gritty -- mostly nits but a few things need a bit of cleaning up. Love the level of detail though!

docs/interpreting-experiments/confidence-intervals.md
bkfunk: Except we use "Significance level" in the product:
image
image

I agree with this point, but as long as we're using significance level in the product, I think it's more helpful to use the language in the actual product itself.

docs/interpreting-experiments/confidence-intervals.md
bkfunk: How should we refer to a metric that has green/red highlighting? We need some way to refer to that, so that we can explain what is happening. I'm happy to use another term than "statistically significant", but it seems like it's easier to at least use a term people will have heard of before than to create a new one

docs/interpreting-experiments/confidence-intervals.md
bkfunk: I put them in the order they are discussed, and I discussed fixed-sample first because it's A) the one people are most likely to have prior experience with, and B) the advantages of sequential (i.e. being able to make a decision at any time) only really make sense when you understand the fixed-sample case. I tried putting sequential first, anyway, but I found it was a lot to discuss the basics of the test (i.e. comparing as if identical) and also highlight the ability to make a decision at any point.

docs/interpreting-experiments/confidence-intervals.md
schmit: Yup unfortunately we might have to use significance level until we reword that in the app, but we don't have to talk about hypothesis testing is my point, and using confidence level instead of significance level merely a trick to help you reword

docs/interpreting-experiments/confidence-intervals.md
schmit: Something like this maybe:

> The confidence interval method generates a confidence interval that contains the true effect with probability at least 95% (or sequence of...), thereby suggesting values outside the confidence interval are implausible. As a special case, we color the confidence interval green or red (depending on desired change) if the confidence interval does not contain zero; or that there is no difference between treatment and control.

Potentially we could add something like:
> Those familiar with hypothesis testing will recognize the correspondence between the confidence interval not containing zero and rejecting the null hypothesis that there is no difference in effect between treatment and control.

schmit:
Note from 2U: We should explicitly define risk and probability beats MDE in the docs.

evanmiller (approved):
tl;dr but we can fix problems later

schmit (approved):
Agree with @evanmiller , this is a massive improvement and we can fix things in follow up PRs if needed

docs/experiments/interpreting-experiments/lift-estimates-and-confidence-intervals/statistical-nitty-gritty.md
```diff
@@ -0,0 +1,560 @@
...
+\begin{equation}
+\begin{align*}
+\lim_{n_C \to \infty}m_C \sim&~~ \mathcal{N}(\mu_C; \frac{\sigma^2_C}{n_C})
+\\[1em]
+\lim_{n_T \to \infty}m_T \sim&~~ \mathcal{N}(\mu_T; \frac{\sigma^2_T}{n_T})
+\end{align*}
+\end{equation}
+$$
+
+:::info Ratio metrics
```
bkfunk, line 78: @schmit I won't block merge on this, but when you get a chance would love your 👀 to make sure I didn't mess anything up in translation from your version. I would really love a good reference for being able to do this normal ratio magic, where we approximate away the cauchy/t even though num. and denom. are correlated here

docs/experiments/interpreting-experiments/lift-estimates-and-confidence-intervals/statistical-nitty-gritty.md
```diff
@@ -0,0 +1,560 @@
...
+\\[1em]
+\hat{\sigma}^2_{\tiny{\Delta}} =&~~ \frac{m_T^2}{m_C^2}\left(
+  \frac{s^2_C}{m_C^2} + \frac{s^2_T}{m_T^2}
+\right)
+\end{align}
+$$
+
+Note that the calculation of the variance $\hat{\sigma}^2_{\tiny{\Delta}}$ relies on the
+[delta method](https://en.wikipedia.org/wiki/Delta_method).
+
```
bkfunk, line 177: @bkfunk TODO: add note about minimum requirements here when that's done, and that we rely on big N for all these normality approx.

PR #63, opened 2022-09-23
Editing Data Warehouse Connecting Docs

hanson377:
Since we have made a massive upgrade to our warehouse connection flows, we need to update our docs.


carlineng (approved):
LGTM!

carlineng:
My only suggestion would be to change `Service User` in the Snowflake section to `service User` (lowercase `s`). Having it capitalized might confuse some users that `Service User` is a specific concept within Snowflake, when it's actually just a `User` that is dedicated to our service.

hanson377:
Makes sense to me, Carlin. I'll change that now.

hanson377:
Does anyone have any idea as to if this just deploys automatically? Or is there something I need to do to deploy it myself?

hanson377:
Oh! Nevermind. Looks like the results are live.

PR #98, opened 2023-01-04
Uninlined the notes

Nickp-eppo:
This is not rendering properly on the site. Still not sure what causing previous netlify errors. Lets see how this goes.

This is a real small change but hopefully fixes rendering issues on the site.


Nickp-eppo:
Oh neat! This deploys to a preview. The preview does indeed look correct :)

PR #101, opened 2023-01-06
update ios sdk to new repo name

leoromanovsky:


leoromanovsky:
verified with netlify test app

PR #130, opened 2023-04-18
remove required data page

lukasgoetzweiss:
I don't think the [required data](https://docs.geteppo.com/experiments/connecting-your-data/assignment-tables/required-data) in our docs is adding much value, and it's a bit misleading (e.g., `device_id` is listed as required)


schmit (approved):

schmit:
Agreed!

PR #141, opened 2023-05-04
Fix broken Experiment Assignment links

williebsweet:
There were three broken links on the [Experiment assignment](http://localhost:3000/experiments/prerequisites/experiment-assignment) page (see below):
- Logging assignment data using Eppo's SDKs
- Exporting assignment data from Launch Darkly
- Exporting assignment data from Optimizely

This PR fixes them. 

You'll may continue to have trouble with your structure and deep links since you are leveraging autogeneration in your sidebar.

### DIFF
--- a/docs/experiments/prerequisites/experiment-assignment/index.md
+++ b/docs/experiments/prerequisites/experiment-assignment/index.md
@@ -24,8 +24,8 @@ It's ok for this table to contain duplicate rows for the same subject, however f
 
 Refer to our tool specific guides to understand how to log assignments to your warehouse:
 
-- [Logging assignment data using Eppo's SDKs](./eppo/)
-- [Exporting assignment data from Launch Darkly](./launch-darkly)
-- [Exporting assignment data from Optimizely](./optimizely)
+- [Logging assignment data using Eppo's SDKs](experiment-assignment/eppo/)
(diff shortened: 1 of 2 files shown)


williebsweet:
I ran this locally, so I know it compiled and ran.
I don't think my change would have effect on the deployment - you may need to re-run it. Of if the PR did do something, just let me know what you see in the logs and I can fix it.

williebsweet:
I looked at the logs.
I think the issue might stem from the use of the 'autogenerated' type in sidebars.js structure. 

The current production version works when navigating from the sidebar to "Experiment Assignment" then clicking the link; however, it breaks when navigating from ["Your first experiment analysis"](https://docs.geteppo.com/experiment-quickstart) > clicking the "experiement assignment" link, and then clicking the links.

The fix proposed in the PR works the opposite way. I assume that's why Netlify failed - it works on the indirect path, not the direct path.

schmit:
Hi @williebsweet , thanks for the PR (sorry just noticed it). These links currently do seem to work for me; do you still see issues on your end?

williebsweet:
@schmit You're right. It is fixed.

Looks the re-org of that section in [this PR](https://github.com/Eppo-exp/eppo-docs/pull/147/files#diff-f3c8e9f380f60bb244c5acb4c05e152d43dbc713accc78be116c531a178f3f7d) fixed it.

PR #147, opened 2023-05-19
Move competitors to reference

lukasgoetzweiss:
Doing some re-organization to push our competitors (LaunchDarkly, Optimizely) to the reference section.

I also restructured the layout of the experimentation section. The most notable changes are simplifying the prerequisites subsection and moving DWH connection into that subsection

PR #152, opened 2023-05-25
Add redirects for recently moved DWH connection docs

lukasgoetzweiss:
In a recent [reorganization of our docs](https://github.com/Eppo-exp/eppo-docs/pull/147), I moved our warehouse connection docs to a different location. This breaks the "Getting Started" tab links on warehouse connection guides. 

I haven't tried using Netlify's `_redirect` file before, but this is the suggestion I found [here](https://github.com/facebook/docusaurus/issues/2843#issuecomment-636863919)

Probably a bit of a blunt instrument, but hopefully will fix the Getting Started page links and also give a better experience for current POCs that may have saved the old URL I've sent to them. I'm open to more elegant solutions if anyone has some, specifically for the second issue


lukasgoetzweiss:
Seems to be working fine on the deploy preview. @leoromanovsky looks like you used `_redirects` for the SEO thing. Any reason for me not to merge this?

PR #166, opened 2023-06-21
Feature flags common issues

sameerank:


docs/feature-flags/common-issues.md
```diff
@@ -0,0 +1,111 @@
+# Common issues
```
sameerank, line 1: Do we need a 

```
---
sidebar_position: 3
---
```
here? The other sections (e.g. Data Management) are explicit about the ordering of the pages, but this isn't the case in the Feature Flags section

schmit: I figured let's get all the pages ready first, and then figure out how to order them :)

docs/feature-flags/common-issues.md
```diff
@@ -0,0 +1,111 @@
...
+
+## 2. Getting assignments via a network
+
+Another example of benefitting from placing assignments close to the experience is avoiding network calls. In addition to batching, it may be tempting to centralization assignment fetching on the server. This may seem especially practical if you don’t pass the the full user object to the front-end and the user attribute needed for the subject key is only available on the backend. However, doing the assignments server-side risks stressing and incurring costs on your network, while our client-side SDKs are designed to get assignments without network calls. Therefore, Eppo recommends for the majority of use cases that the client-side SDK is used whenever possible. There are however a few uses where the server-side SDKs are a better choice for getting assignments, specifically when the experience indicated by the assignment variant originates from the backend, e.g. A/B testing machine learning models whose predictions are generated server-side.
```
emetelka, line 65: "centralization" should be "centralize"

PR #200, opened 2023-09-08
Add REST API page

chasdevs:
Screenshot 2023-09-08 at 11 31 19

Changes:
- Add REST API page
- Remove unused screenshots
- Update api keys page images and text


emetelka (approved):
Looks good. We'll need to update this when Unlimited Environments releases because we'll split SDK keys from REST API keys, but that's on me and Varun to do at that time.

PR #203, opened 2023-09-28
Sameeran/docs 18 get typed assignments

sameerank:


docs/feature-flags/sdks/go.md
```diff
@@ -120,8 +120,9 @@ The `getStringAssignment` function takes two required and one optional input to
 Additional functions are available:
 
 ```
-getBoolAssignment(...)
-getNumericAssignment(...)
+GetBoolAssignment(...)
+GetNumericAssignment(...)
+GetJSONStringAssignment(...)
```
sameerank, line 125: I couldn't find any function for getting parsed JSON

https://github.com/Eppo-exp/golang-sdk/blob/8dc51f3df2598b49b9ec47e0dbfe0e49dec82472/eppoclient/client.go#L36-L53

Should we add it in .. or forget about it until someone asks for it?

Also looks like JSON methods are missing in dot net:

https://github.com/Eppo-exp/dot-net-server-sdk/blob/35c9f0b3f464c43650c9b0f7d5ef1d982d7857ec/dot-net-sdk/EppoClient.cs#L32-L46

docs/feature-flags/sdks/node.md
```diff
@@ -101,21 +101,32 @@ const variation = eppoClient.getAssignment(
...
+```
 
 ### Handling `null`
 
 We recommend always handling the `null` case in your code. Here are some examples illustrating when the SDK returns `null`:
 
 1. The **Traffic Exposure** setting on experiments/allocations determines the percentage of subjects the SDK will assign to that experiment/allocation. For example, if Traffic Exposure is 25%, the SDK will assign a variation for 25% of subjects and `null` for the remaining 75% (unless the subject is part of an allow list).
 
-2. If you are using Eppo for experiment assignments, you must start the experiment in the user interface before `getAssignment` returns variations. It will return `null` if the experiment is not running, both before and after.
+2. If you are using Eppo for experiment assignments, you must start the experiment in the user interface before `getStringAssignment` returns variations. It will return `null` if the experiment is not running, both before and after.
```
sameerank, line 127: This paragraph doesn't make sense to me. You can get assignments as long as the feature flag environment is enabled. You would continue to get assignments even if an experiment is not in a running state.

Am I reading this wrong?

docs/feature-flags/sdks/node.md
```diff
@@ -101,21 +101,32 @@ const variation = eppoClient.getAssignment(
...
 - `flagOrExperimentKey` - This key is available on the detail page for both flags and experiments.
-- `targetingAttributes` - An optional map of metadata about the subject used for targeting. If you create rules based on attributes on a flag/experiment, those attributes should be passed in on every assignment call.
+- `subjectAttributes` - An optional map of metadata about the subject used for targeting. If you create rules based on attributes on a flag/experiment, those attributes should be passed in on every assignment call.
+
+### Typed assignments
+
+Additional functions are available:
+
+```
+getBoolAssignment(...)
```
sameerank, line 115: Sometimes "Bool", sometimes "Boolean" 😑

docs/feature-flags/sdks/node.md
schmit: I'm confused too; I think this may refer to the "Eppo randomization" world prior to building out feature flags.

docs/feature-flags/sdks/node.md
schmit: Is it too late to rename functions?

schmit: We can add the `getBooleanAssignment(...)` function everywhere and add a deprecation warning to the `getBoolAssignment` for example, and remove it in a future version of the SDK

schmit (approved):

docs/feature-flags/dynamic-config.md
```diff
@@ -28,4 +28,18 @@ After you have a control variation, you can create additional variations based o
...
+
+```
+getStringAssignment(...)
+getBoolAssignment(...)
+getNumericAssignment(...)
+getJSONStringAssignment(...)
+getParsedJSONAssignment(...)
+```
+
+The function names differ slightly according to naming conventions in the respective SDK languages. The function used to get the assignment must match the type of the feature flag. To get assignments for a JSON-type feature flag being used as a dynamic config, for example, you would use `getJSONStringAssignment` or `getParsedJSONAssignment`, whereas incorrectly calling `getNumericAssignment` would return `null` and no assignments would occur.
```
aarsilv, line 45: ❤️

docs/feature-flags/sdks/go.md
aarsilv: I think your approach of filing low-priority tickets to eventually do this is a good one.

docs/feature-flags/sdks/node.md
aarsilv: Yeah given the type in JavaScript is "boolean" this is probably the (low priority) path we should take.

sameerank: Added to Linear. This would be nice to do, perhaps we can wrap this into the next round of SDK updates

https://linear.app/eppo/issue/FF-1002/getbool-to-getboolean

docs/feature-flags/sdks/node.md
aarsilv: I recall seeing something that the user can roll out the winning variation. If so, the experiment would still be "enabled" but "ended"? Is this true and if so should we update the docs to reflect this?

aarsilv (approved):
Thanks for going through this and dealing with all the nuances and filing follow-up tickets for irregularities you spotted!

docs/feature-flags/sdks/go.md
sameerank: Sounds good. Added it to the triage: https://linear.app/eppo/issue/FF-1003/golang-missing-getparsedjsonassignment

docs/feature-flags/sdks/node.md
sameerank: I changed it to a sentence about making sure the environment is enabled to avoid getting null assignments. That much makes sense to me

PR #215, opened 2023-10-20
document assignment logging deduplication (FF-1073)

leoromanovsky:
Screenshot 2023-10-20 at 12 09 14 PM

Screenshot 2023-10-20 at 12 08 40 PM

### DIFF
--- a/docs/feature-flags/sdks/index.md
+++ b/docs/feature-flags/sdks/index.md
@@ -20,9 +20,13 @@ will store duplicate data and increase associated costs unnecessarily.
 
 Eppo's SDK deduplicates assignment events using an internal cache
 that takes into account changes in subject variations and other factors to ensure
-one and only one canonical event is transmitted.
+one and only one canonical event is transmitted. 
 
-This behavior is enabled by default and no action is required on your behalf to enable this.
+Since your client likely only handles a single subject, memory pressure should not be a concern 
+and the cache attempts to store all flags for the subject; 
+our goal is to completely eliminate duplicate events.
+
+This behavior is enabled by default and no action is required beyond upgrading to the supported version.
 
 *This functionality is available in the Javascript and React Native clients.*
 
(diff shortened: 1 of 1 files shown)


docs/feature-flags/sdks/index.md
```diff
@@ -12,6 +12,20 @@ Eppo's client SDK integrates with client-side applications that run on a user de
...
 
+### Removing duplicate assignment logs
+
+Each invocation of the `get*Assignment` methods triggers the logging callback function. 
+If you have connected it to transmit events to your Data Warehouse this 
+will store duplicate data and increase associated costs unnecessarily.
+
+Eppo's SDK deduplicates assignment events using an internal cache
+that takes into account changes in subject variations and other factors to ensure
+one and only one canonical event is transmitted.
```
schmit, line 23: This is not quite true right: we don't ensure only 1 event is transmitted, but rather drastically reduce the number of duplicate events

schmit (approved):
Nice! Left one nit

docs/feature-flags/sdks/index.md
leoromanovsky: Ok I'll write it in more words but the goal is only one event is sent only when it needs to be; this might mean "duplicate" but for example if a variation changes 5 times then we need to know that.

schmit: Yup yup agreed. I'm just saying that there might not be a guarantee that there is only 1 unique message sent (e.g. what if the second request goes to a different server and hence that one does not have the cache entry)

leoromanovsky: not to be pedantic but this is the client section so we do expect to almost completely eliminate duplicates here 😄

schmit: Hah okay move my comment to the server side SDK section 😂

PR #219, opened 2023-10-31
Mutually exclusive how to docs

hhargreaveseppo:


docs/how-tos/mutually-exclusive-experiments.md
```diff
@@ -0,0 +1,52 @@
...
+| Feature Flag 2 | copy_test | copy_control | copy_treatment |
+| Feature Flag 3 | layout_test | layout_control | layout_treatment |
+
+::: note
+
+Follow the quickstart [here](/quick-starts/feature-flag-quickstart.md) for additional guidance on how to set up an eppo Feature Flag in the app.
+
+:::
+
+The SDK setup would look like this in your application:
```
lukasgoetzweiss, line 26: What do you think of doing this through targeting rules instead? Something like this:

```
const exclusion_group = eppoClient.getStringAssignment(
  userId,
  "mutual_exclusion_group"
)

copy_variation = eppoClient.getStringAssignment(
		 userId,
		 "copy_test",
                 {"exclusion_group": exclusion_group}
	)

layout_variation = eppoClient.getStringAssignment(
		 userId,
		 "layout_test",
                 {"exclusion_group": exclusion_group}
	)

```

We could add screenshots of what the targeting rules should look like. (The copy test would use `exclusion_group = 'copy_test'`, etc.)

This has two advantages:
1. ability to call each test independently, make rollout decisions without code changes
2. avoid overusing the variable name `variation` (right now it's unclear how the subsequent code should use variation)

wdyt?

docs/how-tos/mutually-exclusive-experiments.md
```diff
@@ -0,0 +1,52 @@
+---
+sidebar_position: 7
+---
+
```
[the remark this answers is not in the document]
lukasgoetzweiss: (needed to make sidebar name look correct)

docs/how-tos/mutually-exclusive-experiments.md
```diff
@@ -0,0 +1,52 @@
...
+There are of course some use cases where it makes sense to make tests mutually exclusive such as when the changes of one experiment effects the experience of another. Take the following scenario for example: the design team wants to run a page redesign test and the marketing team wants to run an experiment on messaging on that same page, and the layout effectively removes the messaging altogether or makes it unreadable. In this situation clearly the changes of one experiment would render the other unusable.
+
+If you do find yourself running an experiment where there would be a true interaction effect, we recommend setting up nested feature flags. Using the same homepage example where each experiment has a control and one treatment, you would create 3 feature flags total where 1 would split traffic into one experiment and the other would split traffic into the other experiment.
```
lukasgoetzweiss, line 11:
```suggestion
If you indeed need to run a mutually exclusive experiment, we recommend setting up nested feature flags. Using the same homepage example, you would create three feature flags: one to split traffic across experiments and one to split traffic for each experiment.
```

lukasgoetzweiss (changes requested):

PR #222, opened 2023-11-08
adding initial mee setup docs

hhargreaveseppo:


docs/how-tos/mutually-exclusive-experiments.md
```diff
@@ -0,0 +1,84 @@
...
+Follow the quickstart [here](/feature-flag-quickstart/) for additional guidance on how to set up an Eppo Feature Flag in the app.
+
+:::
+
+## Eppo App Setup
+
+For your first flag, your setup should look like this:
+
+![Screenshot 2023-11-06 at 3.51.28 PM.png](/img/how-tos/mutually-exclusive-experiments/exclusion-group-flag-setup.png)
+
```
lukasgoetzweiss, line 34: Can we add a sentence (and maybe screenshot) showing how to randomize the `exclusion group`  flag itself? Just to make it clear that there is another layer of randomization here.

docs/how-tos/mutually-exclusive-experiments.md
```diff
@@ -0,0 +1,84 @@
...
+const copy_variation = eppoClient.getStringAssignment(
+												 user.userId,
+												 "copy_test",
+				                 {"exclusion_group": exclusion_group}
+											 )
+// randomize assignment on what variation the user will receive for the Layout Test
+const layout_variation = eppoClient.getStringAssignment(
+												 user.userId,
+												 "layout_test",
+				                 {"exclusion_group": exclusion_group}
```
lukasgoetzweiss, line 82: would it work to just pass in `user`? That way if there are any additional targeting attributes they would be included here

```suggestion
				                 user
```

lukasgoetzweiss (commented):
Looks good! Two small things to consider and then feel free to merge this!

docs/how-tos/mutually-exclusive-experiments.md
hhargreaveseppo: In this particular case, I feel that spelling out what the attribute is to highlight what is happening is important. Once people get it in their own environment I feel confident that they will be savvy enough to understand that they can add additional attributes and refactor

docs/how-tos/mutually-exclusive-experiments.md
hhargreaveseppo: The purpose of the first flag to randomize selection into the next 2 flags is mentioned in the untitled intro paragraph. I'll go ahead and leave out for now, but I'm happy to update/reword if we get feedback.

PR #236, opened 2023-12-19
Fix: Typos

datakabeta:
I came across some typos while using your documentation and figured I might as well quickly fix them instead of raising a support ticket.

I have fixed some typos across 4 files. That is all.


docs/quick-starts/experiments/analysis-plans
schmit, line 1: Not sure how this file got added, but I imagine it was not intended :)

datakabeta: Sorry! I think I know, but I'll save you the back story and delete it right away.

emetelka:
Thanks for the contribution!

datakabeta:
You're welcome!

PR #246, opened 2024-01-11
Minor suggestions

datakabeta:
I am suggesting the following minor improvements. Please feel free to disagree.

1. In the instructions for setting up BigQuery, the _eppo_output_ dataset is created in step 1 and referenced as such in step 2 so it makes sense to reference it as such in step 3. 
2. Per [this comment](https://github.com/Eppo-exp/js-client-sdk-common/blob/0898d84072fb22a67ff659be15c2f44e84ea61c6/src/client/eppo-client.ts#L110) in the JS SDK, the _getAssignment_ function has been deprecated so I have replaced it with _getStringAssignment_. We could though instead use  _get < Type > Assignment_ and add a note to explain 'Type' further.


giorgiomartini0:
Good catches on both @datakabeta , and thank you for the improvements!

PR #249, opened 2024-01-19
Noting python sdk client is initialized async

hhargreaveseppo:


vpai (commented):
sorry im a little late. lgtm

PR #253, opened 2024-01-25
Giant re-org of content

emetelka:


docs/how-tos/_category_.json
```diff
@@ -1,5 +1,5 @@
 {
-    "label": "How-tos",
-    "position": 4
+    "label": "Guides",
```
schmit, line 2: We should rename the directory from `how-tos` to `guides` rather than just renaming the label to avoid getting into a very convoluted state

PR #256, opened 2024-02-01
new first experiment quick start

lukasgoetzweiss:


docs/quick-starts/feature-flag-quickstart.md
```diff
@@ -1,15 +1,17 @@
 ---
 slug: /feature-flag-quickstart
-sidebar_position: 3
+sidebar_position: 2
 ---
 
-# Your first feature flag
+# Your first experiment
```
schmit, line 6: I am not a big fan of this change in naming, I think it makes the nav items very confusing:

Screenshot 2024-01-31 at 4 49 35 PM

lukasgoetzweiss: What do you think about this:

- Launching an experiment 
- Creating a metric
- Analyzing an experiment

I think the content in "Initial setup" could be rolled into the other sections

emetelka: Agree with Sven it's confusing as is. "Launching" or "Deploying" an experiment is better.

PR #259, opened 2024-02-02
Added count distinct explanation

emetelka:


docs/data-management/metrics/simple-metric.md
```diff
@@ -72,6 +81,14 @@ $\frac{\text{COUNT of fact values}}{\text{Number of unique entities assigned}}$
 
 Examples: videos watched per user, articles viewed per visitor, orders per user.
 
+#### Count Distinct
+
+`Count distinct` computes the number of unique non-numeric values in a fact. This allows for counting a number of unique values in a field other than the entity. If the fact value is NULL, it is discarded.
+
+$\frac{\text{Number of unique fact values}}{\text{Number of unique entities assigned}}$
+
+Examples: number of unique devices with a video watch, number of unique cookies who viewed an article, number of unique emails who entered checkout.
```
giorgiomartini0, line 90: The examples feel a bit artificial to me. And the last one especially (unique emails who entered checkout) I don't understand: do users (the entity) have multiple emails? I'm not sure how it would be different from the simple `COUNT`.

Suggested change, based on my understanding of typical use cases for `COUNT DISTINCT`:

```suggestion
Examples: number of unique videos watched per user (if the same video is watched twice, it only counts once), number of unique articles viewed per visitor, number of unique items ordered (if an item is ordered multiple times, it only counts once).
```

PR #272, opened 2024-02-13
new FF architecture page

lukasgoetzweiss:


docs/sdks/overview.md
```diff
@@ -0,0 +1,23 @@
...
+
+In some situations it's not practice to call `getAssignment` right at the code split. For instance, when experimenting on content it may be desirable to pre-fetch content before the user reaches that section of the page. In these situations, it is fine to call `getAssignment` ahead of time and then filter to exposed users on the analysis side. This is supported in Eppo by creating [Entry Points](/experiment-analysis/filter-assignments-by-entry-point).
+
+Several of our SDKs will cache when a flag has already been evaluated for a given user and, if so, not call the logging callback function. Eppo handles duplicated assignment records in the analytics engine, this is simply to help minimize event volume when possible.
```
schmit, line 23: `Several of our SDKs`

Imo this is weak: why not all? and which ones do and which ones don't?

docs/sdks/overview.md
```diff
@@ -0,0 +1,23 @@
...
+
+For more information on Eppo SDK performance see the FAQs on [latency](/sdks/faqs/latency) and [risk](/sdks/faqs/risk).
+
+## Event capture
+
+Eppo's SDK is designed to encourage best practices in event capture. We leverage your existing logging infrastructure to avoid us ever handling your raw data. This is implemented by providing a logging callback function at SDK initialization. The callback function will only be invoked when `getAssignment` is called. Since `getAssignment` is evaluated locally in a few clock cycles, it is encouraged to put the call as close to the code split as possible. This helps capture clean data for analysis: only subjects (users) that were exposed to the experiment will be included.
+
+In some situations it's not practice to call `getAssignment` right at the code split. For instance, when experimenting on content it may be desirable to pre-fetch content before the user reaches that section of the page. In these situations, it is fine to call `getAssignment` ahead of time and then filter to exposed users on the analysis side. This is supported in Eppo by creating [Entry Points](/experiment-analysis/filter-assignments-by-entry-point).
```
schmit, line 21: Might want to add a comment that they do need to log when the content is actually shown

bertilhatt: “by creating [Entry Points] when the user can see the differentiated content.”?

docs/sdks/overview.md
bertilhatt: I’m guessing local memory issues? TBC with devs—if that’s the case, they probably store the x latest, which could make sense to specify here.

lukasgoetzweiss: Afaik, we just haven't had the bandwidth to add it to all of our SDKs. Open to other ways to frame that for this page

lukasgoetzweiss: I'll just remove this paragraph for now

PR #282, opened 2024-02-21
Adding more engineer specific content to SDK section

emetelka:


docs/feature-flagging/01-feature-gates.md
```diff
@@ -1,5 +1,5 @@
 ---
-sidebar_position: 1
+sidebar_position: 2
```
schmit, line 2: nit, these pages are now numbered `01-feature-gates.md` etc, but the sidebar position got shifted

schmit (approved):

docs/feature-flagging/targeting.md
```diff
@@ -1,7 +1,3 @@
----
-sidebar_position: 2
```
lukasgoetzweiss, line 2: are we defining the sidebar_position somewhere else now? I figure targeting should be one of the first ones on the side panel

PR #283, opened 2024-02-21
Add funnel metrics not supported in sample size calculator

MaloChevillotte:


giorgiomartini0 (approved):
Thanks for the quick fix and the additional cleanup!

PR #305, opened 2024-03-13
Added user_id to query

tbuffington7:
It looks like the final result of the query (which I believe is the mapping query) is missing `user_id`


schmit (approved):
good catch!

PR #312, opened 2024-03-21
Hh/node sdk revamp

hhargreaveseppo:


docs/guides/testing-feature-flags.md
```diff
@@ -0,0 +1,33 @@
...
+
+![Screenshot of allocation for Staging Environment](/img/guides/testing-feature-flags/qa-flag-staging-allocation.png)
+
+## Testing in Production
+
+To test in a Production environment, switch the environment on the flag to the **Production Environment**. 
+
+Decide what will indicate an internal user based on what information is available. Create a [Feature Gate](/feature-flagging/feature-gates#create-a-feature-gate) and set up a targeting condition that targets internal users. 
+
+In the example below, we target all users with an internal email address that ends in `@geteppo.com` or a specific user id.
```
lukasgoetzweiss, line 23: Small nit here: I've seen some customers tell their engineers not to pass in PII info like email to our SDK. While we don't record that data, the event capture system will. It might be worth calling that out here in a note. Something like:

:::note
While Eppo does not store any raw data, note that targeting attributes will be passed into the logging callback function. If you do not want your logger to store user emails, you can apply the regex match before calling Eppo and then pass in a simple `is_internal_user` flag to the SDK call.
:::

I'm curious your thoughts, this might be overkill idk.

docs/sdks/server-sdks/node.md
```diff
@@ -31,40 +31,11 @@ npm install @eppo/node-server-sdk
 </TabItem>
 </Tabs>
 
-## 2. Initialize the SDK
```
lukasgoetzweiss, line 34: What do you think of wrapping sections 1-5 in a Getting Started section? 

So the right sidebar would look like this:

```
Getting Started
  Install the SDK
  Create logger
  Initialize the SDK
  Getting variants
  Example
Handling NULL
Typed assignments
Initialization options
```

docs/sdks/server-sdks/node.md
```diff
@@ -31,40 +31,11 @@ npm install @eppo/node-server-sdk
...
-| **`numInitialRequestRetries`** (number) | Number of _additional_ times the initial configurations request will be attempted if it fails. This is the request typically synchronously waited (via `await`) for completion. A small wait will be done between requests. | `1` |
-| **`pollAfterFailedInitialization`** (boolean) | Poll for new configurations even if the initial configurations request failed. | `false` |
-| **`throwOnFailedInitialization`** (boolean) | Throw an error (reject the promise) if unable to fetch initial configurations during initialization. | `true` |
-| **`numPollRequestRetries`** (number) | If polling for updated configurations after initialization, the number of additional times a request will be attempted before giving up. Subsequent attempts are done using an exponential backoff. | `7` |
-
-### Define an assignment logger (experiment assignment only)
+## 2. Define an assignment logger
 
 If you are using the Eppo SDK for experiment assignment (i.e randomization), pass in a callback logging function to the `init` function on SDK initialization. The SDK invokes the callback to capture assignment data whenever a variation is assigned.
```
lukasgoetzweiss, line 36: I've always found this sentence confusing. You are probably initializing the SDK once at app startup. Wouldn't you generally want to use the SDK for both feature gates and experiments? 

I think we should always have them pass in a logger. We've discussed adding more flexibility into when we do or don't invoke the logger that we can add later once fully supported

Let's instead use this section to show that BYO-logger is a good thing, not a feature gap. Something like:

Eppo encourages centralizing application logging as much as possible. Accordingly, instead of implementing a new logging framework, Eppo's SDK integrates with your existing logging system via a logging callback function defined at SDK initialization.

docs/sdks/server-sdks/node.md
```diff
@@ -31,40 +31,11 @@ npm install @eppo/node-server-sdk
...
 
-The code below illustrates an example implementation of a logging callback using Segment. You could also use your own logging system, the only requirement is that the SDK receives a `logAssignment` function. Here we define an implementation of the Eppo `IAssignmentLogger` interface containing a single function named `logAssignment`:
+The code below illustrates an example implementation of a logging callback to Segment. You could also use your own logging system, the only requirement is that the SDK receives a `logAssignment` function. Here we define an implementation of the Eppo `IAssignmentLogger` interface containing a single function named `logAssignment`:
 
 ```javascript
```
lukasgoetzweiss, line 40: Can we do a similar `<Tabs>`  thing like we do in the "Install the SDK" section? And then default to the console logger from the example below:

```
const assignmentLogger: IAssignmentLogger = {
  logAssignment(assignment) {
        console.log(assignment)
  },
};
```

lukasgoetzweiss: We can also add other examples from this page: https://docs.geteppo.com/sdks/event-logging

docs/sdks/server-sdks/node.md
```diff
@@ -84,6 +55,9 @@ const assignmentLogger: IAssignmentLogger = {
 };
 ```
 
+
+
+
 The SDK will invoke the `logAssignment` function with an `assignment` object that contains the following fields:
 
 | Field                     | Description                                                                                                              | Example                             |
```
lukasgoetzweiss, line 63: This table feels a bit like clutter. Can we just mention that Eppo's SDK creates an assignment analytic event and then point the reader either to the example section below, or our [page on event logging](https://docs.geteppo.com/sdks/event-logging)?

docs/sdks/server-sdks/node.md
```diff
@@ -104,8 +80,26 @@ More details about logging and examples (with Segment, Rudderstack, mParticle, a
...
 
+## 3. Initialize the SDK
+
+Initialize the SDK with a SDK key, which can be generated in the Eppo interface. Initialization should happen when your application starts up to generate a singleton client instance, once per application lifecycle:
+
+```javascript
+import { init } from "@eppo/node-server-sdk";
+
+await init({
+  apiKey: "<API_KEY>",
```
lukasgoetzweiss, line 91:
```suggestion
  apiKey: "<SDK_KEY>",
```
It's unfortunate that the code calls it `api_key` instead of `sdk_key`, but maybe this would make it a bit clearer?

docs/sdks/server-sdks/node.md
```diff
@@ -104,8 +80,26 @@ More details about logging and examples (with Segment, Rudderstack, mParticle, a
 
 Eppo's SDK uses an internal cache to ensure that duplicate assignment events are not logged to the data warehouse. While Eppo's analytic engine will automatically deduplicate assignment records, this internal cache prevents firing unnecessary events and can help minimize costs associated with event logging. 
 
+## 3. Initialize the SDK
+
+Initialize the SDK with a SDK key, which can be generated in the Eppo interface. Initialization should happen when your application starts up to generate a singleton client instance, once per application lifecycle:
```
lukasgoetzweiss, line 85:
```suggestion
Initialize the SDK with an SDK key, which can be generated [in the Eppo interface](https://eppo.cloud/feature-flags/keys). Initialization should happen when your application starts up to generate a singleton client instance, once per application lifecycle:
```

docs/sdks/server-sdks/node.md
```diff
@@ -104,8 +80,26 @@ More details about logging and examples (with Segment, Rudderstack, mParticle, a
...
+
+await init({
+  apiKey: "<API_KEY>",
+  assignmentLogger,
+});
+```
+
+After initialization, the SDK begins polling Eppo’s API at regular intervals to retrieve the most recent experiment configurations such as variation values and traffic allocation. The SDK stores these configurations in memory so that assignments thereafter are effectively instant. For more information, see the [architecture overview](/sdks/overview) page.
+
+If you are using the SDK for experiment assignments, make sure to pass in an assignment logging callback (see [section](#define-an-assignment-logger-experiment-assignment-only) below).
```
lukasgoetzweiss, line 98:
```suggestion
```
let's remove this line

docs/sdks/server-sdks/node.md
```diff
@@ -156,3 +192,27 @@ We recommend always handling the `null` case in your code. Here are some example
 :::note
 It may take up to 10 seconds for changes to Eppo experiments to be reflected by the SDK assignments.
```
lukasgoetzweiss, line 193: does this feel a bit out of place?

lukasgoetzweiss: maybe we drop it, we already give that heads up in an in-app modal

docs/sdks/server-sdks/node.md
```diff
@@ -156,3 +192,27 @@ We recommend always handling the `null` case in your code. Here are some example
...
+
+Use the following getAssignment method for the type of feature flag created:
+
+```
+getBoolAssignment(...)
+getNumericAssignment(...)
+getJSONStringAssignment(...)
+getParsedJSONAssignment(...)
+```
+
```
lukasgoetzweiss, line 206:
```suggestion
To read more about different flag types, see the page on [Flag Varitions](/feature-flagging/flag-variations).
```

lukasgoetzweiss (changes requested):
Very big fan of the new direction! I think we just need to make a few small tweaks and then we can publish

docs/sdks/server-sdks/node.md
hhargreaveseppo: I agree -- updated this

docs/sdks/server-sdks/node.md
hhargreaveseppo: updated

docs/sdks/server-sdks/node.md
hhargreaveseppo: i did have that originally, but wasn't sure if it'd be overkill since it's linked to this other doc. I'll add in the tabs for examples here

hhargreaveseppo: updated

docs/sdks/server-sdks/node.md
hhargreaveseppo: moved to the bottom of the page

docs/sdks/server-sdks/node.md
hhargreaveseppo: updated

docs/sdks/server-sdks/node.md
hhargreaveseppo: dropped it

PR #314, opened 2024-03-23
Add Bandit Documentation

aarsilv:
🎟️  **Ticket:** [FF-1777 - Update documentation for bandits](https://linear.app/eppo/issue/FF-1777/update-documentation-for-bandits)

Here we add in the documentation about interacting with bandits to the Java Server SDK.

I also added a higher-level bandits section in the SDK Features section. Note it uses Java examples (many from the Java Server SDK documentation), as this is the only SDK we have right now.

### DIFF
--- a/docs/sdks/server-sdks/java.md
+++ b/docs/sdks/server-sdks/java.md
@@ -69,6 +69,118 @@ The SDK will invoke the `logAssignment` function with an `event` object that con
 More details about logging and examples (with Segment, Rudderstack, mParticle, and Snowplow) can be found in the [event logging](/sdks/event-logging/) page.
 :::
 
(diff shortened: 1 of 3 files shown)


docs/sdks/server-sdks/java.md
```diff
@@ -69,6 +69,118 @@ The SDK will invoke the `logAssignment` function with an `event` object that con
 More details about logging and examples (with Segment, Rudderstack, mParticle, and Snowplow) can be found in the [event logging](/sdks/event-logging/) page.
 :::
 
+### Define a bandit assignment logger (contextual multi-armed bandit assignment only)
```
schmit, line 72: I think it makes more sense to move the bandit section of the Java SDK to the bottom so that all the feature flag content is grouped and the bandit content is grouped separately

PR #315, opened 2024-03-25
Updated ratio metric docs to explain motivation and provide more deta…

tbuffington7:
…iled use-cases

Our current ratio metrics documents are a bit lacking in detail. IMO, they don't equip users with sufficient knowledge to understand whether a given metric needs to be a simple metric vs. a ratio metric.

There is even more confusion because _not all ratios are ratio metrics_, and simple metrics are ratios (they normalize by the count of assigned entities).

After several discussions, we decided that the best way to explain this topic is to introduce examples that clearly articulate the defining characteristics of ratio metrics. The ARPU/ARPPU one is especially interesting because average revenue per assigned user is a simple metric, but average revenue per paying user is a ratio metric. This distinction is especially effective in communicating the difference.

### DIFF
--- a/docs/data-management/metrics/ratio-metric.md
+++ b/docs/data-management/metrics/ratio-metric.md
@@ -3,14 +3,53 @@ sidebar_position: 2
 ---
 
 # Ratio metrics
+## Overview
+Ratio metrics allow you to calculate the ratio of two [simple metrics](/data-management/metrics/simple-metric). The most common use case for this is when the randomization entity is different from the analysis entity. 
+For example, consider the metric _average  order value_. Typically, experiments are not randomized at the _order_ level because a user can have multiple orders. 
+Consequently, randomizing orders could lead to users experiencing multiple variants. When the experiment is randomized at the user level, it is not statistically valid to consider each
+_order_ an independent observation because orders from the same user are related. Treating the orders as statistically independent observations in experiments with user-based
+randomization typically leads to overconfident results (underestimating the variance). A proper estimation of the variance must recognize the fact that the observations (orders) are clustered
+at the user level. Although there are several valid approaches for estimating this variance, including [clustered standard errors](https://en.wikipedia.org/wiki/Clustered_standard_errors),
+Eppo uses the approach described in [Deng et al.](https://alexdeng.github.io/public/files/kdd2018-dm.pdf) and recommended in 
+Chapter 18 of _Trustworthy Online Controlled Experiments: A Practical Guide to A/B Testing_ by Ron Kohavi.
 
-Ratio metrics allow you to calculate ratio of two [simple metrics](/data-management/metrics/simple-metric), providing deeper insights into the relative performance of variations. This enables a more nuanced analysis of experimental results, allowing businesses to understand the impact of changes in a more comprehensive manner.
 
-For example, consider an _average order value_ metric, which is created by dividing revenue (sum of prices) by number of orders (_sum_ of items purchased or _count_ of prices).
+The approach is to redefine the metric of interest as a ratio of two simple metrics. For example, take _average order value_, which is defined by 
+dividing revenue (sum of prices) by the number of orders (sum of items purchased or count of prices).
+
 
 $$
 \text{Average order value} = \frac{\text{Revenue}}{\text{Number of orders}}
 $$
+
+We can multiply both the numerator and the denominator by the number of _users_ in the experiment:
+
+$$
+\text{Average order value} = \frac{\text{(Revenue)(Number of users)}}{\text{(Number of orders)(Number of users)}}
+$$
+
+which we can rearrange as follows:
+$$
+\text{Average order value} = \bigg(\frac{\text{Revenue}} {\text{Number of users}}\bigg) \bigg(\frac{\text{Number of users}} {\text{Number of orders}}\bigg)
+$$
+
+which can be expressed as the ratio of two simple metrics:
+$$
+\text{Average order value} = \frac{\text{Revenue per user}}{\text{Number of orders per user}}
+$$
+
+Both the numerator and the denominator are simple metrics whose analysis entity matches the randomization entity (a user). In general, the steps to define a ratio metric are
+as follows:
+1. Define the metric you wish to calculate and identify its analysis entity (e.g., an order in _average order value_).
+2. Divide both the numerator and the denominator by the _randomization_ entity (often a user).
+3. Identify the numerator metric and the denominator metric. The numerator metric will typically be a fact quantity per randomization entity (e.g., revenue per user). The denominator metric will typically be a count of the analysis entity per randomization entity (e.g., number of orders per user).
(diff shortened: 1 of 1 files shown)


docs/data-management/metrics/ratio-metric.md
```diff
@@ -3,15 +3,54 @@ sidebar_position: 2
 ---
 
 # Ratio metrics
+## Overview
+Ratio metrics allow you to calculate the ratio of two [simple metrics](/data-management/metrics/simple-metric). The most common use case for this is when the randomization entity is different from the analysis entity. 
+For example, consider the metric _average  order value_. Typically, experiments are not randomized at the _order_ level because a user can have multiple orders.
```
schmit, line 8:
```suggestion
For example, consider the metric _average order value_. Typically, experiments are not randomized at the _order_ level because a user can have multiple orders. 
```

schmit (approved):
Awesome, thanks for updating!

docs/data-management/metrics/ratio-metric.md
```diff
@@ -3,15 +3,54 @@ sidebar_position: 2
...
+$$
+\text{Average order value} = \frac{\text{Revenue per user}}{\text{Number of orders per user}}
+$$
+
+Both the numerator and the denominator are simple metrics whose analysis entity matches the randomization entity (a user). In general, the steps to define a ratio metric are
+as follows:
+1. Define the metric you wish to calculate and identify its analysis entity (e.g., an order in _average order value_).
+2. Divide both the numerator and the denominator by the _randomization_ entity (often a user).
+3. Identify the numerator metric and the denominator metric. The numerator metric will typically be a fact quantity per randomization entity (e.g., revenue per user). The denominator metric will typically be a count of the analysis entity per randomization entity (e.g., number of orders per user).
+4. Make sure that both the numerator metric and the denominator metric are already defined as [simple metrics](/data-management/metrics/simple-metric) in Eppo.
```
lukasgoetzweiss, line 46: You shouldn't need to create separate numerator and denominator simple metrics before creating a ratio metric. As long as the fact is defined, you can go straight to creating a ratio metric

docs/data-management/metrics/ratio-metric.md
```diff
@@ -3,15 +3,54 @@ sidebar_position: 2
...
+$$
+
+which can be expressed as the ratio of two simple metrics:
+$$
+\text{Average order value} = \frac{\text{Revenue per user}}{\text{Number of orders per user}}
+$$
+
+Both the numerator and the denominator are simple metrics whose analysis entity matches the randomization entity (a user). In general, the steps to define a ratio metric are
+as follows:
+1. Define the metric you wish to calculate and identify its analysis entity (e.g., an order in _average order value_).
```
lukasgoetzweiss, line 43: Should this say "define the fact you wish to ..." 

I also wonder if using the word "entity" here could lead to some confusion (since that is another concept in Eppo, just not something that we currently connect back to this use case)

docs/data-management/metrics/ratio-metric.md
```diff
@@ -3,15 +3,54 @@ sidebar_position: 2
...
+2. Divide both the numerator and the denominator by the _randomization_ entity (often a user).
```
lukasgoetzweiss, line 44: Eppo will automatically divide by the randomization entity, I think reads a bit like the user has to define this themselves

docs/data-management/metrics/ratio-metric.md
```diff
@@ -3,15 +3,54 @@ sidebar_position: 2
...
+3. Identify the numerator metric and the denominator metric. The numerator metric will typically be a fact quantity per randomization entity (e.g., revenue per user). The denominator metric will typically be a count of the analysis entity per randomization entity (e.g., number of orders per user).
```
lukasgoetzweiss, line 45: I think it's easier to conceptualize in the context of defining the ratio metric aggregation. I.e., in the numerator you'll choose "sum of X" and in denominator you'll choose "count of x"

lukasgoetzweiss (changes requested):

docs/data-management/metrics/ratio-metric.md
tbuffington7: Good catch! The first line of this documentation--"Ratio metrics allow you to calculate ratio of two [simple metrics](https://docs.geteppo.com/data-management/metrics/simple-metric)"--made me think that we need to configure the numerator and denominator as simple metrics, but now I see that when you actually create a ratio metric, it's done directly from the facts, not from existing simple metrics.

Addressed your comment here: https://github.com/Eppo-exp/eppo-docs/commit/d3a2614bda7e7ab9ce81d9479f15c1c36f8054b9

docs/data-management/metrics/ratio-metric.md
tbuffington7: Yeah, good point. I first learned about calculating AOV (and ratio metrics generally) from Ron Kohavi's HiPPO book, which frames the calculation as a ratio of two averages, so that's always felt more intuitive to me (probably because it's easier to see that averages are normally distributed by CLT, so the delta method feels more natural). However, for most users, thinking of the numerator as a sum of X and the denominator as the count of x is more straightforward in light of Eppo's terminology.

I updated it here, but it could still use some polishing: https://github.com/Eppo-exp/eppo-docs/commit/a1982eee5c58d85e6981dae8e1d0f2e87ea73a53

docs/data-management/metrics/ratio-metric.md
tbuffington7: Good callout. I made relevant updates here https://github.com/Eppo-exp/eppo-docs/commit/a1982eee5c58d85e6981dae8e1d0f2e87ea73a53, however based on our [Slack thread](https://eppo-group.slack.com/archives/C058QMZ6XQC/p1711399886470789) I don't feel confident that I fully understand how this is done. Let's sync in our 1x1

docs/data-management/metrics/ratio-metric.md
```diff
@@ -3,15 +3,36 @@ sidebar_position: 2
...
 
 $$
 \text{Average order value} = \frac{\text{Revenue}}{\text{Number of orders}}
 $$
 
+This metric is often calculated in experiments that are randomized at the user level. However, it is clear from this metric's definition that we need to normalize by the
+number of **orders** rather than the number of **users** (which simple metrics would do automatically). To capture this custom aggregation logic in the denominator, we define it as a ratio metric.
+The numerator aggregation is the sum of revenue, and the denominator aggregation is the count of orders.
+
+### Example 2: Average revenue per paying user (ARPPU)
```
tbuffington7, line 21: @emetelka I know you mentioned that you've typically seen this acronymized as ARPU, but I think having the extra "P" here for "paying" makes it clearer for folks like me who come from an industry where ARPU stands for "average revenue per user" and it includes all assigned users. Happy to change it if you disagree, though

emetelka (approved):

lukasgoetzweiss (approved):

PR #317, opened 2024-03-27
Create a guide section for Entry points

bertilhatt:
I wanted to explain why one might want to use an Entry point and give a few examples of when that pattern is relevant.

Remain to be done:
* [x] Include links in the manual page
* [ ] Graphical illustration
  * [ ] timeline for each example
  * [ ] illustrations?
* [ ] Ask for other examples in the #cds channel


docs/guides/entry_points.md
```diff
@@ -0,0 +1,90 @@
...
+
+More generally, complex models, some API calls, and AI services are non-trivial to scale and **expensive to serve**. You might want to limit the cost of testing Treatment by not wasting half of it. In a gradual roll-out process, you want organise your feature flag so that only internal developpers and testers see the results first, then expose a small minority of visitors to the A/B test, then scale it to a full 50/50 test. That allows you to control who you expose, and if Treatment is expensive, how much valuable ressources you dedicate to testing an idea.
+
+During internal tests, it matters less, but once you run an A/B test on users, if you want properly validate the impact of those changes, an Entry point will help you get better, sharper results.
+
+### More examples
+
+There are many more example where the assignement has to happens before entities can see a difference, like API calls during time-sensitive ad auctions; or where the different treatment is triggered, but might not be visible to users, like including legal warning on a pop-up that could be blocked by the browser.
```
lukasgoetzweiss, line 88:
```suggestion
There are many more examples where the assignment has to happen before entities can see a difference, like API calls during time-sensitive ad auctions; or where the different treatment is triggered, but might not be visible to users, like including a legal warning on a pop-up that could be blocked by the browser.
```

lukasgoetzweiss (changes requested):

docs/guides/entry_points.md
```diff
@@ -0,0 +1,90 @@
...
+
+If you A/B test a better customer service, you’ll end up splitting the 2,000 who call into 1,000 for Control (about 250 of which should give you a good rating) and 1,000 for Treatement. If the new service is a lot better and three quarters of customers calling you now give you a good rating, that’s 250/1000 vs. 750/1000, i.e. a very clear result. Experiments are noisy so it won’t be exactly 250 and 750, but probably 250±15 and 750±15.
+
+You can likely make a decision that the new approach to customer service is better within a few weeks.
+
+If you include all your customers in the process, then you’ll have 9,000 extra participants in each variant that are not affected but the test, half of which will rate the service highly. Score will be around 4,500 + 250 = 4,750/10,000 vs. 5,250/10,000. The new treatment is still better but results will be noisier, probably around 4,750±100 and 5,250±100. The result after one month might not be conclusive. Your decision might have to wait for more evidence, while all you need is to focus on the information you already have.
```
emetelka, line 23:
```suggestion
If you include all your customers in the process, then you’ll have 9,000 extra participants in each variant that are not affected but the test, half of which will rate the service highly. The score will be around 4,500 + 250 = 4,750/10,000 vs. 5,250/10,000. The new treatment is still better but results will be noisier, resulting in a wider confidence interval of ±100. The result after one month might not be conclusive. Your decision might have to wait for more evidence, while all you need is to focus on the information you already have.
```

docs/guides/entry_points.md
```diff
@@ -0,0 +1,90 @@
...
+
+In this case too, the assignment happens when confirming basket contents and the address, but the entry point should be set to a little later, when the customer sees a quote.
+
+::: note
+
+If one of the API integration fails, and some customers assigned to Treatment don’t see the pricing quote because of a bug, then you’ll observe an [asymetric traffic](/statistics/sample-ratio-mismatch.md).
+
+:::
+
+There are more cases where tests on your deliveries would need an Entry point. Let’s say you want to improve the box opening experience, include a note, etc. The assignment has to hapen when packaging; the entry point has to be set when the customer receives the note. There’s no reason to include customers who didn’t get their packages when measuring the impact of a hand-written note on their satisfaction.
```
emetelka, line 64:
```suggestion
There are more cases where tests on your deliveries would need an Entry Point. Let’s say you want to improve the box opening experience, include a note, etc. The assignment has to happen when packaging; the Entry Point has to be set when the customer receives the note. There’s no reason to include customers who didn’t get their packages when measuring the impact of a hand-written note on their satisfaction.
```

docs/guides/entry_points.md
```diff
@@ -0,0 +1,90 @@
...
+
+Not all logistical test require a separate Entry point. If you want to test a promotion that offers free delivery under certain condition, you can decide which customer sees that promotion when it becomes visible, not earlier.
```
emetelka, line 66:
```suggestion
Not all logistical test require a separate Entry Point. If you want to test a promotion that offers free delivery under certain condition, you can decide which customer sees that promotion when it becomes visible and not earlier.
```

docs/guides/entry_points.md
```diff
@@ -0,0 +1,90 @@
...
+
+::: note
+
+Be mindful when testing promotions on anonymous users: you risk contaminating your experiment. If they get re-assigned by refreshing their browser, or opening a private window, some users will stumble on it, notice the discrepancy, and figure out the trick. They might leak that information. If prospective buyers see it, they will try. You are better off testing promotions on entities that buyers can’t change easily: established user account, phone number, postal address.
```
emetelka, line 70: I think this is outside the bounds of this guide and should be removed. We're hypothesizing abut a bunch of risks we don't have real knowledge about

docs/guides/entry_points.md
```diff
@@ -0,0 +1,90 @@
...
+
+:::
+
+### Third example: AI-generated customer service
+
+Finally, let’s say you want to test one of those chat-bots that promise to replave human-lead customer service. The vendor recommends to send them as much information about the customer as you can, including their queries and the product decription that they open during their session. Small issue: user logs are streamed to your data warehouse, but you don’t really store that information in the browser memory or a convenient session object that could be sent it as single blob to the AI company. No problem: the vendor can ingest page content as the visitor goes through your site, store them and will use those as context when a customer asks a question.
```
emetelka, line 76:
```suggestion
Finally, let’s say you want to test an AI chatbots for customer service. The vendor recommends to send them as much information about the customer as you can, including their queries and the product description that they open during their session. You accomplish this by sending the vendor page content as the visitor goes through your site, storing that information and using it as context when a customer asks a question.
```
A few sentences about implementation that aren't relevant and can be cut. You just need to establish that user data is sent to the AI bot before a user sees it.

docs/guides/entry_points.md
```diff
@@ -0,0 +1,90 @@
...
+
+Once again, you have to decide to send information about a customer, i.e. assign them to either Control or Treatment before they ask a question and are exposed to the AI bot. 
+
+In this example (as well as the examples about triggering complex recommendations) you could send an API call for **all users** and only decide which response to use once you need to show the information to the users. That pattern can avoid having to use Entry point. However, that vendor might charge per visitor, and you might want to lower the cost of testing their solution.
```
emetelka, line 80: The vendor charging per user is hypothesizing that we don't need to include. Let's focus on concrete examples as much as possible. If you want to walk through an example of minimizing cost per user, you should carve out that example explicitly.

docs/guides/entry_points.md
```diff
@@ -0,0 +1,90 @@
...
+
+More generally, complex models, some API calls, and AI services are non-trivial to scale and **expensive to serve**. You might want to limit the cost of testing Treatment by not wasting half of it. In a gradual roll-out process, you want organise your feature flag so that only internal developpers and testers see the results first, then expose a small minority of visitors to the A/B test, then scale it to a full 50/50 test. That allows you to control who you expose, and if Treatment is expensive, how much valuable ressources you dedicate to testing an idea.
```
emetelka, line 82:
```suggestion
More generally, complex models, some API calls, and AI services are non-trivial to scale and **expensive to serve**. You might want to limit the cost of testing Treatment by not wasting half of it. In a gradual roll-out process, you want organise your feature flag so that only internal developers and testers see the results first, then expose a small minority of visitors to the A/B test, then scale it to a full 50/50 test. That allows you to control who you expose, and if Treatment is expensive, how much valuable resources you dedicate to testing an idea.
```

docs/guides/entry_points.md
```diff
@@ -0,0 +1,90 @@
...
+
+During internal tests, it matters less, but once you run an A/B test on users, if you want properly validate the impact of those changes, an Entry point will help you get better, sharper results.
+
+### More examples
```
emetelka, line 86: I would just cut this section

docs/guides/entry_points.md
```diff
@@ -0,0 +1,90 @@
...
+
+There are many more example where the assignement has to happens before entities can see a difference, like API calls during time-sensitive ad auctions; or where the different treatment is triggered, but might not be visible to users, like including legal warning on a pop-up that could be blocked by the browser.
+
+If you have any questions, don’t hesitate to ask your Eppo contact.
```
emetelka, line 90: Reach out to [Eppo support](mailto:support@geteppo.com)

docs/guides/entry_points.md
```diff
@@ -0,0 +1,90 @@
+---
+sidebar_position: 11
+---
+
+# Filtering assignments by entry point
```
emetelka, line 5:
```suggestion
# When to add Entry Point filters
```

docs/guides/entry_points.md
```diff
@@ -0,0 +1,90 @@
...
+
+There’s one concern though: while all visitors will be assigned in the experiment, only the visitors who see the recommendation carousel are exposed to a different experience. (We’ll ignore the impact of triggering an expensive computation for now.) Therefore, the two thirds of visitors who were assigned but never scrolled down the homepage should not be included in that experiment.
+
+For cases like that, we let you define an **Entry Point**: what event needs to happen for visitors to be exposed to a different experience, and considered enrolled to the experiment. It remains up to you to decide if this should be when the carousel enters the viewport, is fully or partially visible; it’s also up to your front-end developpers to trigger and log that event. But once that information is in your data warehouse, then you can use it to filter out which users participate in the experiment.
+
+::: note
+If you define an Entry point, all the time-limted metrics (“Conversion 7 days after assignment”) are based on the timestamp of the Entry point, not the assignment.
```
tbuffington7, line 42: I think this language is a little confusing. My understanding was that entry points were a way to specify that the assignment happens based on an entry point, so the timestamp of the entry point is the same as the timestamp of the assignment. Is that incorrect?

docs/guides/entry_points.md
tbuffington7, line 82:
```suggestion
More generally, complex models, some API calls, and AI services are non-trivial to scale and **expensive to serve**. You might want to limit the cost of testing Treatment by not wasting half of it. In a gradual roll-out process, you want to organize your feature flag so that only internal developers and testers see the results first, then expose a small minority of visitors to the A/B test, and then scale it to a full 50/50 test. That allows you to control who you expose and, if Treatment is expensive, how much valuable resources you dedicate to testing an idea.
```

docs/guides/entry_points.md
tbuffington7, line 80: I'm a little confused here... If we assign users to control or treatment **before** they ask a question, why do we need to send an API call for all users? If we know a user is in Control, then we shouldn't need to send any of their information to the vendor, right?

tbuffington7 (changes requested):
Nice examples! Left a few comments and suggested edits

docs/guides/entry_points.md
```diff
@@ -0,0 +1,90 @@
...
+
+You also can easily imagine offering renting a car, or a valuable asset and needing a quote from your insurance company, based on many factors: information about the person renting, time and duration, etc. 
+
+Let‘s say you want to test asking for fast and cheap options. Or you could be testing a new provider. Alternatively, you might want to test asking both for a quote and let customers compate. Whatever is the change you want to test, you will need to decide which API call to trigger before the customer is guaranteed to see pricing because you have to wait for confirmation, say, that the items are in stock.
+
+In this case too, the assignment happens when confirming basket contents and the address, but the entry point should be set to a little later, when the customer sees a quote.
+
+::: note
+
+If one of the API integration fails, and some customers assigned to Treatment don’t see the pricing quote because of a bug, then you’ll observe an [asymetric traffic](/statistics/sample-ratio-mismatch.md).
```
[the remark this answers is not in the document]
bertilhatt: The better option is the client’s existing monitoring of their API calls.  They should have an alert long before an SRM flags it as a problem. I’m trying to pepper some awareness of how SRM happens in a context where it could be relevant.
What I can do is mention that explicitly, so that people don‘t think they can rely on SRM detectors —because they shouldn’t.

docs/guides/entry_points.md
```diff
@@ -0,0 +1,90 @@
...
+
+The visitors who don’t see a difference should behave the same between Control and Treatment, so they likely won’t skew your results one way on another. But having more unaffected customers will add **noise** to your results and make it harder to test a good idea apart. You often hear that more participants is better; that’s only true if they are actually participating in the test. More unaffected subjects won’t help.
+
+### Numeric Example 
+
+Let’s say you have 20,000 subscriber coming to your side and buying something per month. About half of your users give you good ratings. Only 10% call customer service. About a quarter of those give you good ratings after the call, and are more likely to come back.
+
+If you A/B test a better customer service, you’ll end up splitting the 2,000 who call into 1,000 for Control (about 250 of which should give you a good rating) and 1,000 for Treatement. If the new service is a lot better and three quarters of customers calling you now give you a good rating, that’s 250/1000 vs. 750/1000, i.e. a very clear result. Experiments are noisy so it won’t be exactly 250 and 750, but probably 250±15 and 750±15.
```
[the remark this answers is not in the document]
bertilhatt: Good catch. I shouldn’t try to do math in my head.

docs/guides/entry_points.md
bertilhatt: Let me clarify that part.

docs/guides/entry_points.md
```diff
@@ -0,0 +1,90 @@
...
+
+You can likely make a decision that the new approach to customer service is better within a few weeks.
+
+If you include all your customers in the process, then you’ll have 9,000 extra participants in each variant that are not affected but the test, half of which will rate the service highly. Score will be around 4,500 + 250 = 4,750/10,000 vs. 5,250/10,000. The new treatment is still better but results will be noisier, probably around 4,750±100 and 5,250±100. The result after one month might not be conclusive. Your decision might have to wait for more evidence, while all you need is to focus on the information you already have.
+
+## Examples when an Entry point is useful
```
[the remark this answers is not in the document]
bertilhatt: @emetelka seems attached to capitalizing key concepts of our interface.

docs/guides/entry_points.md
bertilhatt: Removed that detail

docs/guides/entry_points.md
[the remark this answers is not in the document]
bertilhatt: Rephrased with the math to help people follow

docs/guides/entry_points.md
bertilhatt: Removed

docs/guides/entry_points.md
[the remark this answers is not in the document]
tbuffington7: > @emetelka seems attached to capitalizing key concepts of our interface.

That's fine with me; my objection is to inconsistent capitalization and mixed capitalization ("Entry point" in the middle of a sentence).

It should either be "entry point" everywhere (except the start of a sentence) or "Entry Point" everywhere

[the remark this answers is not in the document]
bertilhatt: Gotcha. Fixed

tbuffington7 (approved):
Looks good from my side. Thanks for grinding through all the comments so quickly!

lukasgoetzweiss (approved):

PR #322, opened 2024-03-29
Bandit quickstart

emetelka:


docs/quick-starts/bandit-quickstart.md
```diff
@@ -0,0 +1,188 @@
...
+slug: /bandit-quickstart
+sidebar_position: 5
+---
+
+# Running your first Contextual Bandit
+
+This 10 minute guide will walk through configuring and running your first Contextual Bandit with Eppo. Note that Contextual Bandits leverage the Eppo SDK and this guide will walk through SDK setup.
+
+:::info
+In order to optimize a Bandit, you must have [created your first metric](/metric-quickstart).
```
bertilhatt, line 11: Would a reader know that “optimize” here means “run” a Bandit, or could they assume it means to improve an existing Bandit setup?

docs/quick-starts/bandit-quickstart.md
```diff
@@ -0,0 +1,188 @@
...
+
+For now, create a Test environment SDK key by using the "New SDK Key" button. Give the key a name and select "Test" for the Environment.
+
+![Generate a SDK key](/img/feature-flagging/sdk-key-modal.png)
+
+Store the SDK key securely; it is not possible to view it after closing the modal. However, generating a new key is easy in case you do lose it.
+
+### 2. Connect a logging function to the Eppo SDK
+
+Eppo leverages your existing event logging infrastructure to track experiment assignment. Whether you are using a third party system to log events to the data warehouse or have an internally built solution, you'll simply pass in a logging function when initializing the SDK.
```
bertilhatt, line 28:
```suggestion
Eppo leverages your existing event logging infrastructure to track experiment assignments. Whether you use a third-party system to log events to the data warehouse or have an internally built solution, you'll simply pass in a logging function when initializing the SDK.
```

docs/quick-starts/bandit-quickstart.md
```diff
@@ -0,0 +1,188 @@
+---
+slug: /bandit-quickstart
+sidebar_position: 5
+---
+
+# Running your first Contextual Bandit
+
+This 10 minute guide will walk through configuring and running your first Contextual Bandit with Eppo. Note that Contextual Bandits leverage the Eppo SDK and this guide will walk through SDK setup.
```
bertilhatt, line 8:
```suggestion
This 10-minute guide will walk through configuring and running your first Contextual Bandit with Eppo. Note that Contextual Bandits leverage the Eppo SDK and this guide will walk through SDK setup.
```

docs/quick-starts/bandit-quickstart.md
```diff
@@ -0,0 +1,188 @@
...
+
+Store the SDK key securely; it is not possible to view it after closing the modal. However, generating a new key is easy in case you do lose it.
+
+### 2. Connect a logging function to the Eppo SDK
+
+Eppo leverages your existing event logging infrastructure to track experiment assignment. Whether you are using a third party system to log events to the data warehouse or have an internally built solution, you'll simply pass in a logging function when initializing the SDK.
+
+The [event logging](/sdks/event-logging/) page has more information on how to set up logging using different logging tools
+
+This logger should write to a table with the following columns (they can be in any order):
```
bertilhatt, line 32: Assuming the names have to to exactly those:
```suggestion
This logger should write to a table with columns with the following names (they can be in any order):
```
Otherwise, 
```suggestion
This logger should write to a table with columns with the following information. They can be in any order and have different names, but you will need to configure them to let Eppo know which is which:
```

docs/quick-starts/bandit-quickstart.md
```diff
@@ -0,0 +1,188 @@
...
+```
+
+If the actions don't have attributes for the context, you can simply provide a set of actions without attributes:
+```java
+Set<String> actions = Set.of("dog", "cat", "bird", "goldfish");
+
+Optional<String> banditAssignment = eppoClient.getStringAssignment(subjectKey, flagKey, subjectAttributes, actions);
+```
+
+Note that the `getAssignment` methods in Eppo are deterministic, meaning that they will always return the same variant for a given subject (e.g., user) for the life of the bandit.
```
bertilhatt, line 169: That's probably surprising to many people familiar with Bandit. I’m not sure if this would need a clarification like, “When Bandit decides which options are preferable, new it will expose  subjects to those options.” 
You might also need to explain: “Once Bandit has converged on the best option, you will be able to A/B test and roll out that option to all subjects.” for our clients who have mostly permanent users and don’t want most of them to get stuck with sub-optional options.

docs/quick-starts/bandit-quickstart.md
```diff
@@ -0,0 +1,188 @@
...
+
+### 7. Train the Bandit 
+:::note
+Check that you're in the Test environment
+:::
+
+Using the template provided for Bandit Training, click `Add Feature Gate Allocation`. We are going to serve the Bandit to 100% of traffic in the test environment. Click the `Save` button to confirm.
+
+![Add bandit training to the Test environment](/img/bandit/bandit-qs-3.png)
```
bertilhatt, line 178: Splitting the bandit in two, with one half compared to Control and the other for the rest of the traffic, is counterintuitive.

The visible separation on this screen doesn’t help the reader understand that it’s the same Bandit, training on the same data.

More importantly, the reason to separate Control with a sliver of Bandit in a balanced portion (even if it’s a pattern used elsewhere by Eppo, to account for an incremental roll-out without containing initially excluded users from Control they might later leave) isn’t the most intuitive. We might have to explain somewhere why we don’t present a single bar with 5% Control and 95% Bandit. That would feel closer to how we’d present it.

bertilhatt (approved):
Great introduction, only minor suggestions

docs/quick-starts/bandit-quickstart.md
schmit, line 169: > for the life of the bandit.

This is not true: the action can change for the user after we update the model parameters (otherwise, how can the bandit be adaptive). However, if you call the bandit twice in quick succession for the same user and with the same context/actions, it will return the same action.

docs/quick-starts/bandit-quickstart.md
schmit, line 178: Agree with Bertil's comment above but also in the testing environment it makes no sense to set up both allocations. Better to simply turn on the bandit (training) only and avoid the extra complexity about analysis for this moment.

docs/quick-starts/bandit-quickstart.md
```diff
@@ -0,0 +1,188 @@
...
+
+### 8. Turn on the Bandit
+
+To start the Bandit, flip the flag on in the Test environment.
+
+![Turn on the bandit using the toggle](/img/bandit/bandit-qs-4.png)
```
schmit, line 184: This screenshot should show the toggle flipped, and perhaps pointed out with an arrow

docs/quick-starts/bandit-quickstart.md
```diff
@@ -0,0 +1,188 @@
...
+
+You should now see assignments coming through the Eppo SDK!
+
+To deploy to production, create a new SDK key for the production environment, create a production experiment allocation, and enable the flag.
```
schmit, line 188: I think here is where we add a section that explains how to go to production, and why you want to set up both training and analysis. We can then also point out that they can now set up an experiment to analyze the bandit's performance

schmit (approved):
Overall this is looking great; added a few comments.

docs/quick-starts/bandit-quickstart.md
aarsilv: Specifically, variation assignments are deterministic. If a user is assigned control, they will always get control. If assigned the bandit, they will always get the bandit.

However, as pointed out, with bandits, actions could change over time each time the bandit model is updated (currently daily). Thus, the _output_ of the `getStringAssignment()` is not deterministic when a bandit is involved.

docs/quick-starts/bandit-quickstart.md
aarsilv: Maybe it would help to "defang the snake" and include a parenthetical explaining how in production we'll also be allocating some traffic to a control variation.

docs/quick-starts/bandit-quickstart.md
```diff
@@ -0,0 +1,200 @@
...
+
+![Add bandit training to the Test environment](/img/bandit/bandit-qs-3.png)
+
+### 8. Turn the Bandit on in Test
+
+To start the Bandit, flip the flag on in the Test environment.
+
+![Turn on the bandit using the toggle](/img/bandit/bandit-qs-4.png)
+
+You should now see assignments coming through the Eppo SDK!
```
aarsilv, line 186: Maybe we should say you should "soon" see assignments coming through the Eppo SDK, as polling is done every 30 seconds with some random jitter (Up to 5 seconds for Java SDK) as well.

docs/quick-starts/bandit-quickstart.md
```diff
@@ -0,0 +1,200 @@
...
+
+1. Create a new SDK key for the production environment from the SDK Keys tab
+2. In the Test environment, click the `Duplicate allocation` button next to Bandit Training and choose to copy it to Production
+![Copy the Bandit Training to Production](/img/bandit/bandit-qs-5.png)
+3. Switch to the Production environment using the environment switcher
+4. Using the template provided, click to add an experiment allocation for `Bandit Analysis`. This will allow you to measure the impact of your Bandit on a small percentage of traffic.
+![Production setup](/img/bandit/bandit-qs-6.png)
+5. Toggle the flag on in Production
+
+Congrats on running your first Bandit! As a next step, you can connect your flag to an Experiment Analysis so you can measure the impact your bandit is having on your product. See the [experiment quickstart guide](/experiment-quickstart) for more.
```
aarsilv, line 200: I wonder if we should provide any color on what will happen. For example, until the bandit assignments are analyzed by our daily pipeline, an action will be randomly chosen. Each day, we'll analyze how assignments impacted the metric we're optimizing for, and start weighting action probabilities by the provided context and what we've learned.

aarsilv (commented):
Great stuff! 🎉  Happy to approve once we clarify the "deterministic" nature of `getAssignment()` as it relates to bandits. 👌 

Dropped a bunch of comments you're welcome to use or ignore as you see fit.

docs/quick-starts/bandit-quickstart.md
emetelka: I think that's good in the context of the bandit reference doc, but not the quickstart doc

docs/quick-starts/bandit-quickstart.md
```diff
@@ -0,0 +1,201 @@
...
+Give the Bandit a descriptive human readable name. Select the entity and input the name of table in your warehouse where you are logging bandit actions from Step 2. Choose a metric to optimize the Bandit for and how long you want the lookback window to be. The lookback window defines the time period the Bandit looks for changes in the optimization metric for subjects who were assigned.
+
+![Setup Bandit step](/img/bandit/bandit-qs-1.png)
+
+### 4. Add variations
+To evaluate the impact of the Contextual Bandit, we run an experiment with exactly two variations: the bandit and a control. This control variation should be the "status quo" that you implement on your end. A simple example is a fixed action.
+
+![Add variations step](/img/bandit/bandit-qs-2.png)
+
+### 5. Initialize the SDK
```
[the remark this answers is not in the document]
emetelka: I went back and forth on it. Initialize the SDK is also step 5 in the FF quickstart. But let me try moving it up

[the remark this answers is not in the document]
emetelka: Actually, I think it's fine where it is. I want to minimize the switching between the application and code. To initialize the SDK the logger should be defined. But we also need the table for the logging data to create the bandit. That's why I put logging so high up.

After the logger is defined, the user then creates the bandit, which is needed to embed the bandit. Then the user does SDK initialization and embeds the bandit back to back.

It's definitely tricky. The need for the warehouse table to create the bandit is obstacle to what I think could be a more straightforward flow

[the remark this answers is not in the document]
aarsilv: That is a good point. Doesn't need to be done now, but we could have the logger example be creating a variable (rather than inlined in the client initialization) that is then passed to the initialization like the example in my other comment

PR #325, opened 2024-04-01
Added guide on running clustered experiments

tbuffington7:
Added a draft of a new guide that outlines running clustered experiments.

I'm making this a draft for now because I think we need to clarify what Eppo _can_ and _can't_ accommodate with its current functionality. I think the main limitations are with regard to entry points. I.e. we randomize by company, but we only want to include users in the experiment after they have performed action X.

### DIFF
--- a/docs/guides/analyzing-clustered-experiments.md
+++ b/docs/guides/analyzing-clustered-experiments.md
@@ -35,15 +35,16 @@ address this issue based on the approach described
 in [Deng et al.](https://alexdeng.github.io/public/files/kdd2018-dm.pdf) and recommended in Chapter 18 of _Trustworthy
 Online Controlled Experiments_. The approach is to recognize that a metric defined at a finer grain than the
 randomization unit can be expressed as the ratio of two metrics at the same grain as the randomization unit. To
-illustrate this,
-we will outline two examples.
+illustrate this, we will outline two examples.
 
 ## Example 1: Average Order Value (AOV)
 
 _Average order value_ is one of the examples mentioned in the [Ratio Metrics](/data-management/metrics/ratio-metric)
 documentation. Although we can recognize that it is a Ratio
 Metric because it is normalized by a count of orders rather than a count of assigned users (which would make it a Simple
-Metric), it is also a good example of a clustered metric.
+Metric), it is also a good example of a clustered metric. Understanding how this familiar ratio metric is a clustered
+analysis provides the intuition for how Ratio Metrics accommodate more complicated clustered analyses.
+
 Typically average order value is calculated in experiments that are randomized at the user level, but AOV is a "per
 order" metric. Each user can have multiple orders, so a user is
 a _cluster_ of orders. As a result, we need to make sure that our statistical calculations recognize the user-level
--- a/docs/guides/analyzing-clustered-experiments.md
+++ b/docs/guides/analyzing-clustered-experiments.md
@@ -2,7 +2,7 @@
 sidebar_position: 2
 ---
 
-# Running Clustered Experiments
+# Analyzing Clustered Experiments
 
 ## Overview
 
--- a/docs/guides/analyzing-clustered-experiments.md
+++ b/docs/guides/analyzing-clustered-experiments.md
@@ -42,8 +42,9 @@ illustrate this, we will outline two examples.
 _Average order value_ is one of the examples mentioned in the [Ratio Metrics](/data-management/metrics/ratio-metric)
 documentation. Although we can recognize that it is a Ratio
 Metric because it is normalized by a count of orders rather than a count of assigned users (which would make it a Simple
-Metric), it is also a good example of a clustered metric. Understanding how this familiar ratio metric is a clustered
-analysis provides the intuition for how Ratio Metrics accommodate more complicated clustered analyses.
+Metric), it is also a good example of a clustered metric. This makes this example a particularly helpful introduction
+because customers who use AOV are already running clustered experiment analyses, even if they do not realize it.
+Understanding how this familiar ratio metric is a clustered analysis provides the intuition for how Ratio Metrics accommodate more complicated clustered analyses.
 
 Typically average order value is calculated in experiments that are randomized at the user level, but AOV is a "per
 order" metric. Each user can have multiple orders, so a user is
--- a/docs/guides/analyzing-clustered-experiments.md
+++ b/docs/guides/analyzing-clustered-experiments.md
@@ -1,7 +1,3 @@
----
-sidebar_position: 2
----
-
 # Analyzing Clustered Experiments
 
 ## Overview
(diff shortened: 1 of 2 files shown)


lukasgoetzweiss (approved):
Great writeup!

Is the issue with entry points just that we need some definition of which users correspond to a given company? I.e., if a user has no activity, Eppo would not count them towards the numerator or the denominator.

I can see an argument that in this case you actually want to include those users in the denominator (someone on a call referred to this as "filling in NULLs with 0s"). If there's a table that maps the association between companies and users every day, they could do a count distinct on user_id for that table, which should solve this.

Might be worth adding a "considerations" section that talks through that

tbuffington7:
> Great writeup!
> 
> Is the issue with entry points just that we need some definition of which users correspond to a given company? I.e., if a user has no activity, Eppo would not count them towards the numerator or the denominator.
> 
> I can see an argument that in this case you actually want to include those users in the denominator (someone on a call referred to this as "filling in NULLs with 0s"). If there's a table that maps the association between companies and users every day, they could do a count distinct on user_id for that table, which should solve this.
> 
> Might be worth adding a "considerations" section that talks through that

Thanks! Regarding entry points, I think that's part of it. IIUC, another component is that we can't easily define metrics based on when the analysis unit was exposed to the treatment? For example, if we have a company-randomized experiment and we want to calculate a user-level conversion rate based on a window of "within 7 days of assignment," that won't work at the user level. I think we can only do within 7 days of when the entire company was assigned to the experiment.

docs/guides/running-clustered-experiments.md
```diff
@@ -0,0 +1,130 @@
+---
+sidebar_position: 2
```
schmit, line 2: Do you have to update the sidebar_position for other articles?

schmit (approved):

docs/guides/running-clustered-experiments.md
```diff
@@ -0,0 +1,130 @@
...
+---
+
+# Running Clustered Experiments
```
giorgiomartini0, line 5: I think this guide should be called `Analyzing Clustered Experiments`. The guide explains how to define analysis unit-level metrics in terms of randomization unit-level metrics. Running a clustered experiment requires much more, especially when clusters are not given to you for free (such as example 2 below) and the hard part is to build suitable clusters.

docs/guides/running-clustered-experiments.md
```diff
@@ -0,0 +1,130 @@
...
+
+Fortunately, [Ratio Metrics](/data-management/metrics/ratio-metric)
+address this issue based on the approach described
+in [Deng et al.](https://alexdeng.github.io/public/files/kdd2018-dm.pdf) and recommended in Chapter 18 of _Trustworthy
+Online Controlled Experiments_. The approach is to recognize that a metric defined at a finer grain than the
+randomization unit can be expressed as the ratio of two metrics at the same grain as the randomization unit. To
+illustrate this,
+we will outline two examples.
+
+## Example 1: Average Order Value (AOV)
```
giorgiomartini0, line 45: Leading with this example is IMO not in the spirit of a "guide". It reads more academic: _here's this thing you took for granted, and is actually trickier than you thought; but it still falls under this approach_. The takeaway for an Eppo user is not clear: should they do anything differently?

The second example is more salient, and more clearly explains how to solve a new problem. In fact, I'd like to see more done to highlight the last sentence, and that this is the "how to" part of the guide.

giorgiomartini0 (commented):
Very clear and well-written 👏  Left a few comments on positioning and emphasis.

docs/guides/running-clustered-experiments.md
tbuffington7: Great point. Changed here: https://github.com/Eppo-exp/eppo-docs/commit/09661436891d9853ac23cfe1db1da1cf7def4403

docs/guides/running-clustered-experiments.md
tbuffington7: Good points. Originally, I was hoping it would read more like this: _this approach is actually less tricky than you thought; in fact, you're probably already using it even though you don't realize it_. However, I agree it is a bit academic and should emphasize the "how to" component more.

Here are some ideas to address this:
1. Reverse the order of the examples
2. Remove the AOV example entirely
3. Give more context for the AOV example without changing the order

I'll think more about it...

docs/guides/running-clustered-experiments.md
tbuffington7: I just removed this. I had copied it over from another template https://github.com/Eppo-exp/eppo-docs/commit/51f012a67d16f13f814bed0c7bc9a9bcdc2fa43c

docs/guides/running-clustered-experiments.md
tbuffington7: I added more context to clarify some of these points. See https://github.com/Eppo-exp/eppo-docs/pull/325/commits/04e9a3c3671e342e1fb86b904b2fab540070c7b2 and https://github.com/Eppo-exp/eppo-docs/pull/325/commits/5e2f356b93e49f2fe58697ce846d7419b48c0916.

I made it more explicit that this example is a good introduction because users are likely already running clustered analyses even if they don't realize it. Hopefully this makes it clear that they don't need to do anything different with AOV for the clustered case.

giorgiomartini0: The extra context is helpful, thanks!

docs/guides/analyzing-clustered-experiments.md
```diff
@@ -42,8 +42,9 @@ illustrate this, we will outline two examples.
 _Average order value_ is one of the examples mentioned in the [Ratio Metrics](/data-management/metrics/ratio-metric)
 documentation. Although we can recognize that it is a Ratio
 Metric because it is normalized by a count of orders rather than a count of assigned users (which would make it a Simple
-Metric), it is also a good example of a clustered metric. Understanding how this familiar ratio metric is a clustered
-analysis provides the intuition for how Ratio Metrics accommodate more complicated clustered analyses.
+Metric), it is also a good example of a clustered metric. This makes this example a particularly helpful introduction
+because customers who use AOV are already running clustered experiment analyses, even if they do not realize it.
```
giorgiomartini0, line 46: Minor, but the third person here ("customers", "they") feels off. here feels off to me. It would make sense if the doc were directed at Eppo employees; but it's primarily for the customers themselves.

giorgiomartini0 (approved):

docs/guides/analyzing-clustered-experiments.md
tbuffington7: Good call! Second person feels awkward here too, so I just avoided using a pronoun. https://github.com/Eppo-exp/eppo-docs/commit/6d471c9cb2c79c953f7d0778a0abfb4b3b1c27e3

PR #338, opened 2024-04-15
Update SDK documentation for v3

giorgiomartini0:
~**Ready to review. Do not merge before v3 SDKs are released.**~ Ready to go!

Comprehensive changes to prepare the SDK documentation for v3. This does _not_ contain the migration guide.

This PR only contains changes for Python- and JavaScript-based SDKs. Other languages (and bandits, since those are bsed on Java) are in [a separate PR](https://github.com/Eppo-exp/eppo-docs/pull/339).

The bulk of the change is updating `getStringAssignment` in all the code examples and docstrings. There are a few incidental changes commented inline.


docs/guides/integrating-with-contentful.md
```diff
@@ -68,13 +68,14 @@ await init({
 });
 
 let userid = “user id”
-let flag_key = <EPPO FLAG KEY>
+let flagKey = <EPPO FLAG KEY>
```
giorgiomartini0, line 71: JS does not use snake_case

docs/quick-starts/bandit-quickstart.md
```diff
@@ -164,7 +164,7 @@ If the actions don't have attributes for the context, you can simply provide a s
 ```java
 Set<String> actions = Set.of("dog", "cat", "bird", "goldfish");
 
-Optional<String> banditAssignment = eppoClient.getStringAssignment(subjectKey, flagKey, subjectAttributes, actions);
+Optional<String> banditAssignment = eppoClient.getStringAssignment(subjectKey, flagKey, defaultVariation, subjectAttributes, actions);
```
giorgiomartini0, line 167: @aarsilv heads up that I added defaults for the Java SDK. This will be inaccurate until there is a Java v3, but hopefully doesn't create too much confusion

docs/sdks/client-sdks/android.md
```diff
@@ -77,44 +77,33 @@ More details about logging and examples (with Segment, Rudderstack, mParticle, a
...
-1. The **Traffic Exposure** setting on experiments/allocations determines the percentage of subjects the SDK will assign to that experiment/allocation. For example, if Traffic Exposure is 25%, the SDK will assign a variation for 25% of subjects and `null` for the remaining 75% (unless the subject is part of an allow list).
-
-2. Assignments occur within the environments of feature flags. You must enable the environment corresponding to the feature flag's allocation in the user interface before `getStringAssignment` returns variations. It will return `null` if the environment is not enabled.
-
-![Toggle to enable environment](/img/feature-flagging/enable-environment.png)
-
-3. If `getStringAssignment` is invoked before the SDK has finished initializing, the SDK may not have access to the most recent experiment configurations. In this case, the SDK will assign a variation based on any previously downloaded experiment configurations stored in local storage, or return `null` if no configurations have been downloaded.
-
-<br />
-
```
giorgiomartini0, line 117: Here and many other languages: 🪓 . No more `null`!

docs/sdks/common-issues.md
```diff
@@ -2,7 +2,7 @@
 sidebar_position: 9
 ---
 
-# Best practices
+# Common pitfalls
```
giorgiomartini0, line 5: Renaming this page because the sections are things _not_ to do, not best practices.

docs/sdks/common-issues.md
```diff
@@ -106,15 +106,11 @@ Now, simply add allocation logic specifying to target users with `beta_user = 't
...
-
-This however can lead to issues in analyzing an experiment. First, if some traffic receives a `NULL` assignment due to network issues, they will all land in the control group. This breaks the core assumption of AB experimentation: that on average there is no difference between the two groups other than the feature on which you are experimenting.
+This however can lead to issues in analyzing an experiment. First, if some traffic receives the default assignment due to network issues, they will all land in the control group. This breaks the core assumption of AB experimentation: that on average there is no difference between the two groups other than the feature on which you are experimenting.
 
 Further, if the allocation percentage changes over time, additional bias may be introduced. Imagine ramping up an experiment to 50% after one week of smoke testing at 10%. If you were to randomly select a user from the control group, they are now more likely to be from the first week of the experiment. Again, there is a difference between the two groups (control is more likely to be from the first week).
```
giorgiomartini0, line 115: @schmit I decided to leave the final section in; it talks about something important (ramping up allocation percentage instead of changing variations weights), but the point is harder to understand now that "default assignment" replaces `null`.

docs/feature-flagging/feature-gates.md
```diff
@@ -32,7 +32,7 @@ user_attributes = {
     device: user.device
 }
 
-if (eppoClient.getStringAssignment(user.id, 'new-checkout-page', user_attributes) == 'on'):
+if (eppoClient.getStringAssignment(user.id, 'new-checkout-page', 'off', user_attributes) == 'on'):
```
schmit, line 35: Feels like this makes more sense as a boolean flag:

```suggestion
if (eppoClient.getBoolAssignment(user.id, 'new-checkout-page', false, user_attributes)):
```

docs/feature-flagging/kill-switches.md
```diff
@@ -26,7 +26,7 @@ Here's how you would use the switch in your code:
 ```python
 # The argument user_id is required but doesn't affect assignment in this case.
 
-if (eppoClient.getStringAssignment(user_id, 'twilio-killswitch') == 'on'):
+if (eppoClient.getStringAssignment(user_id, 'twilio-killswitch', 'off') == 'on'):
```
schmit, line 29: Could also turn this into a boolean flag

schmit: Also the code suggests this should be Python code:

```
if eppo_client.get_boolean_assignment(user_id, 'twilio-killswitch', false):
```

docs/guides/integrating-with-contentful.md
schmit: Should we also write `userId`?

docs/feature-flagging/targeting.md
```diff
@@ -40,8 +40,9 @@ import eppo_client
 
 client = eppo_client.get_instance()
 variation = client.get_string_assignment(
-  "<SUBJECT-KEY>", 
-  "<EXPERIMENT-KEY>", 
+  "<SUBJECT-KEY>",
+  "<FLAG-KEY>",
+  "<DEFAULT-VARIATION>",
```
schmit, line 45: This is slightly confusing perhaps: I think it's clearer to say "default value" instead of default variation

docs/guides/mutually-exclusive-experiments.md
```diff
@@ -60,7 +60,8 @@ const eppoClient = EppoSdk.getInstance();
 // randomize assignment for which exclusion group they will be in
 const exclusion_group = eppoClient.getStringAssignment(
   userId,
-  "mutual_exclusion_group"
+  "mutual_exclusion_group",
+  "<default_variation>"
```
schmit, line 64: Same on this page re default variation vs default value

docs/feature-flagging/kill-switches.md
giorgiomartini0: Good catch. I updated to Python syntax, but leaving it as a string flag (1) to match the screenshots, (2) True and False are somewhat ambiguous for a kill switch.

docs/sdks/sdk-features/obfuscation.md
```diff
@@ -45,6 +45,8 @@ The `value` field in `conditions` may or may not be hashed depending on the conf
 | greater than (>)           | encoded |
 | greater than or equal (>=) | encoded |
 | matches regex              | encoded |
+| is null                    | n/a     |
+| is not null                | n/a     |
```
schmit, line 49: Nit: in the SDK this is a single operator where the value indicates true or false, which is hashed

schmit (approved):

docs/quick-starts/bandit-quickstart.md
giorgiomartini0: Or not; I've pulled out these changes into https://github.com/Eppo-exp/eppo-docs/pull/339. We can apply them when the Java SDK is updated to v3.

docs/feature-flagging/feature-gates.md
aarsilv: I still think we should poll the larger group of engineers about these method signatures. This call looks so weird to me to have the default where it is and subject stuff separated. I'd expect something like:
```py
eppoClient.getStringAssignment('new-checkout-page', user.id, user_attributes, 'off' )
```
Of course not all languages make function overloading easy, so my next choice would be:
```py
eppoClient.getStringAssignment('new-checkout-page', 'off', user.id, user_attributes)
```

docs/guides/integrating-with-contentful.md
aarsilv: while in here, these should be `const` and terminate with semicolons 😛 

And maybe make it valid by putting the placeholder inside a string.
```ts
const flagKey = "<EPPO FLAG KEY>";
```

docs/sdks/common-issues.md
aarsilv: Good catch 📈

docs/sdks/common-issues.md
```diff
@@ -106,15 +106,11 @@ Now, simply add allocation logic specifying to target users with `beta_user = 't
...
 
-## 6. Using `NULL` assignments to track control
-
-When you are running an AB experiment, it is best practice to explicitly assign users as control. If you have already confirmed that null values are handled well and serves a default experience (see above), it may be tempting to serve 10% of traffic the new variant, serve the remaining 90% `NULL`, and then analyze the experiment using `NULL` as the control group.
-
-This however can lead to issues in analyzing an experiment. First, if some traffic receives a `NULL` assignment due to network issues, they will all land in the control group. This breaks the core assumption of AB experimentation: that on average there is no difference between the two groups other than the feature on which you are experimenting.
+This however can lead to issues in analyzing an experiment. First, if some traffic receives the default assignment due to network issues, they will all land in the control group. This breaks the core assumption of AB experimentation: that on average there is no difference between the two groups other than the feature on which you are experimenting.
```
aarsilv, line 113: Should we note that when the SDK returns the default assignment, it won't have logged the assignment, so it's excluded from analysis? Otherwise, we may confuse or scare people!

docs/sdks/sdk-features/obfuscation.md
aarsilv, line 1: We will now obfuscate a lot more but probably not worth referencing in the docs

aarsilv (approved):
Approving as-is, but I don't think the parameter ordering case is closed yet, and we should solicit feedback from the larger team / take note of what competing SDKs do.

docs/guides/integrating-with-contentful.md
giorgiomartini0: Good catch @aarsilv ; cleaned this up a bit more.

PR #340, opened 2024-04-16
Fix broken links in SDK documentation

tbuffington7:
I noticed some strange behavior in our docs. If you navigate directly to the Server SDK page (i.e. click on this [link](https://docs.geteppo.com/sdks/server-sdks)), and then scroll down and click on one of the languages, it routes you to a broken link (e.g. https://docs.geteppo.com/sdks/python). 

However, if you navigate to the same page by clicking "Server SDKs" in the sidebar, it works as expected. I suspect this is due to quirks with relative paths. I changed the references to absolute paths in this PR and that seems to fix the issue.

Ditto for Client SDKs.

image

### DIFF
--- a/docs/sdks/client-sdks/index.md
+++ b/docs/sdks/client-sdks/index.md
@@ -18,7 +18,7 @@ The SDK uses a background process to fetch and store the experiment data. The p7
 
 ### Language-specific Documentation
 
-- [JavaScript](./javascript)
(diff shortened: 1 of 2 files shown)


tbuffington7:
The problem seems to have fixed itself in prod

lukasgoetzweiss (approved):

PR #342, opened 2024-04-17
adding deduplicating assignments section

hhargreaveseppo:


docs/data-management/definitions/assignment-sql.md
```diff
@@ -63,3 +63,19 @@ Holdout columns are also defined here. More information is available on the [ded
...
+
+## Handling duplicate assignments
+
+Eppo has built in filtering to handle scenarios where there are multiple assignment events for a subject with different data. Here are a few common scenariios that you may encounter and how Eppo handles them:
+
+**1. Subject has multiple assignment events where they received multiple variations.**
+
+Example: A subject received both the `Control` and `Treatment` within an experiment. 
+
+Eppo will automatically filter these subjects out of that experiment's results.
```
jonathan-fulton, line 75: There's a caveat here.  If the exposures happen within a couple days of each other (incremental lookback period), then we'll filter them out completely.  If the multiple exposures happen outside of that incremental lookback period, then data for the subject is included up until the exposure to another variant.  @chasdevs correct me if I'm wrong here, but I think we dropped the additional join to the assignments table to filter out all data because it was too slow.

lukasgoetzweiss: Looking at Paddle, it seems like the check for `is_mixed` happens during the assignmentSourceCumulative stage, which does not appear to have any incremental logic

i.e.,I think we're removing users who received multiple variants even if it's across a long time period

lukasgoetzweiss (approved):

docs/data-management/definitions/assignment-sql.md
jonathan-fulton: Yes, we're filtering out is_mixed data once the subject is determine to be mixed.  The problem comes if there was a gap larger than the incremental lookback window.  Then the mixed subjects have data in the daily data frame and they're included in the experiment.

jonathan-fulton: The daily data frame is computed incrementally so we don't delete the "is_mixed" data.  At one point we had an additional filtering step where we joined back to the assignments table but Evan's wide table refactor removed that additional filtering.

chasdevs: @jonathan-fulton I think we addressed that issue in this PR: https://github.com/Eppo-exp/eppo/pull/8362

jonathan-fulton: That's great, thanks @chasdevs!

hhargreaveseppo:
Thanks for double/triple checking team, I'm gonna go ahead and merge this now

PR #352, opened 2024-05-02
Hh/no low code guide

hhargreaveseppo:
This will get replaced once Layers is available. This topic has been coming up in enough deals that it's important enough to have stop-gap documentation in the meantime.


emetelka (approved):

docs/guides/no-low-code-experiments.md
```diff
@@ -0,0 +1,101 @@
...
+After adding an additional Experiment Allocation, the allocation shows up on your Feature Flag overview page.
+
+![Second experiment allocation on a flag](/img/guides/no-low-code-experiments/second-exp-allocation.png)
+
+
+This new allocation logic splits the remaining users 50/50 into the second and third variations if those users do not qualify for the first 2 allocations. The previous experiment allocation can be archived by selecting the ‘bin’ icon on the right hand side of the allocation. Additionally, by clicking the gray side bar on the right hand side of the allocation, allocations can be dragged and dropped to reprioritize the order of logic executed.
+
+:::warning
+
+Eppo does not recommend running two experiments on the same flag at the same time since it can interfere with experiment results. We recommend archiving the experiment allocation after the test is complete before adding a new experiment allocation to the same flag. The example above is to illustrate the ability to have multiple experiments created on the same flag.
```
lukasgoetzweiss, line 26: This seems odd to me - the guide is showing how to do it, but then we say we actually don't recommend doing it.

Is the main concern that a user could drift from one experiment to another? Maybe this warning should really say to only run this kind of experiment when users are unlikely to change populations (e.g., if the targeting is based on primary country, etc.)

docs/guides/no-low-code-experiments.md
```diff
@@ -0,0 +1,101 @@
...
+
+:::
+
+You can now run multiple experiments off of this same flag without any additional releases. Continuing this example, this is the code required for this flag regardless of how many experiment allocations or feature gates are created:
+
+```jsx
+import * as EppoSdk from '@eppo/node-server-sdk';
+
+const client = EppoSdk.getInstance();
+const variation = client.getAssignment("winter-promo", "<SUBJECT-KEY>", <SUBJECT-ATTRIBUTES>, "<DEFAULT-VALUE>")
```
lukasgoetzweiss, line 36:
```suggestion
const variation = client.getAssignment(
    "winter-promo", 
    "<SUBJECT-KEY>", 
    <SUBJECT-ATTRIBUTES>, 
    "<DEFAULT-VALUE>"
)
```

is it better to have new lines and indents? (I know this is standard style guide in python, not sure about JS)

docs/guides/no-low-code-experiments.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+
+## Using JSON flags to parameterize frequently tested areas
+
+JSON flags can be used to parameterize your app or site to run experiments without an additional release. By using a [JSON flag](/feature-flagging/flag-variations#json-flags), values from Eppo dynamically update predefined values in your code. Extending the feature flag example above, variables in your code can be created to accept a value defined in the JSON object defined in Eppo.
+
+Below is an example extended from our [React](/sdks/client-sdks/javascript#usage-in-react) documentation.
+
+```
```
lukasgoetzweiss, line 46:
```suggestion
```jsx
```

lukasgoetzweiss (approved):

docs/guides/no-low-code-experiments.md
hhargreaveseppo: Yea, i went back and forth on this--maybe I just make the stated workflow "archive the previous experiment, and now you're running a new experiment on the same flag" and remove the warning.

Last time we went through this, there was a strong opinion to make sure users were aware of the downstream effects of running multiple concurrent experiments on the same flag.

hhargreaveseppo: ended up taking out for now

PR #353, opened 2024-05-03
Multiple allocations

bertilhatt:
Wanted to mention the possibility of having multiple allocations, notably multiple experiments, in the docs


bertilhatt:
By the way, I’ve been describing this as a “flag pole.” I'm not sure the metaphor makes sense, but I think being illustrative helps a bit.

docs/feature-flagging/experiment-assignment.md
```diff
@@ -20,6 +20,10 @@ Click "Create Experiment" on the experiment allocation to create an experiment.
 
 ![Randomization 1](/img/feature-flagging/randomization-1.png)
 
+Each allocation (feature gate or experiment) can target one subset of subjects thanks to rules. Subjects for whom the first rule don’t apply are passed on to the next allocation on the flag. Thanks to this, multiple experiments can be run on the same flag, targetting distinct users.
```
emetelka, line 23:
```suggestion

### Run multiple experiments (optional)
You can add multiple experiment allocations that target different segments of traffic. This can be accomplished by creating multiple experiment allocation with mutually exclusive targeting rules. Subjects for whom the first targeting rule don’t apply are passed on to the next allocation on the flag.

You may want to do this if you want to test different variants for mutually exclusive groups, such as variants that apply to subjects in different countries.
```

PR #362, opened 2024-05-17
Hh/nextjs guide

hhargreaveseppo:


docs/guides/nextjs-setup.md
```diff
@@ -0,0 +1,285 @@
...
+cd react-sdk-walkthrough
+yarn dev
+```
+
+Open a browser to the location output for `local` (e.g., [http://localhost:3000](http://localhost:3000/))
+
+## Eppo flag setup
+1. [Create and copy an SDK key](/sdks/sdk-keys.md) in your Eppo app if you don't already have one.
+2. [Create a feature flag](/quick-starts/feature-flag-quickstart.md/#2-create-a-flag) in your Eppo app.
+3. Make sure your flag is turned on the environment that correlates to your SDK key.
```
schmit, line 33:
```suggestion
3. Make sure your flag is turned on in the environment that correlates to your SDK key.
```

docs/guides/nextjs-setup.md
```diff
@@ -0,0 +1,285 @@
...
+
+### Install the Eppo JS client SDK
+
+```bash
+yarn add @eppo/js-client-sdk
+```
+
+### Create a randomization provider
+
+1. Create a provider based off our [Usage in React](/sdks/client-sdks/javascript#usage-in-react) documentation
```
schmit, line 45: This is confusing: do I need to go to that page? It seems to do the same as what's listed below

docs/guides/nextjs-setup.md
```diff
@@ -0,0 +1,285 @@
...
+
+  if (!waitForInitialization || isInitialized) {
+    return children;
+  }
+  return loadingComponent;
+}
+
+export default EppoRandomizationProvider;
+```
+4. Replace API key placeholder with your SDK key
```
schmit, line 86:
```suggestion
4. Replace `<SDK-KEY>` placeholder with your SDK key
```

docs/guides/nextjs-setup.md
```diff
@@ -0,0 +1,285 @@
...
+import { useMemo } from "react";
+import { getInstance } from "@eppo/js-client-sdk";
+
+const OfferComponent = () => {
+
+    const subjectKey = "<SUBJECT-KEY>"
+
+    const assignedVariation = useMemo(() => {
+      const eppoClient = getInstance();
+      return eppoClient.getStringAssignment("<FLAG-KEY>", subjectKey, <SUBJECT-ATTRIBUTES>, "<DEFAULT-VALUE>");
```
schmit, line 101: I find the inconsistency a bit jarring, maybe do this instead?
```suggestion
      return eppoClient.getStringAssignment(<FLAG-KEY>, subjectKey, <SUBJECT-ATTRIBUTES>, <DEFAULT-VALUE>);
```

docs/guides/nextjs-setup.md
```diff
@@ -0,0 +1,285 @@
...
+        {assignedVariation === "control" && <p>50% off widgets!</p>}
+        {assignedVariation === "test" && <p>Buy one widget get one free!</p>}
+        {!assignedVariation && <p>Buy a widget today</p>}
+      </div>
+    );
+  }
+
+export default OfferComponent;
+```
+In the example above the disabled/error (`null`) case is handled as well as assigned variations.
```
schmit, line 115: Should never be `null` right? Since this is using v3 of the SDK

schmit (approved):
Sweet, added a few nits but looks great!

docs/guides/nextjs-setup.md
hhargreaveseppo: Yea, I copied/pasted from other examples in our docs. I think the difference is conveying this value is a string and this value is not a string. I agree though, maybe all the string quotes should be removed and people can reference the schema for the method.

docs/guides/nextjs-setup.md
hhargreaveseppo: ah good catch, this was written in the before times

docs/guides/nextjs-setup.md
hhargreaveseppo: Good call, going to make that an intro vs a step in the process

PR #368, opened 2024-05-28
Updated query, added test query to verify that there are no overlappi…

tbuffington7:
…ng intervals

- Updated the anon mapping query to partition by user_Id, which is a better default than partitioning by anonymous_id, which can lead to double-counting of facts
- Added a test query to verify that there are no overlapping intervals


lukasgoetzweiss (approved):
LGTM. Only question is how we tested these queries. Is there a synthetic dataset somewhere? If not, do we feel highly on it from the recent examples we've been working on?

tbuffington7:
> LGTM. Only question is how we tested these queries. Is there a synthetic dataset somewhere? If not, do we feel highly on it from the recent examples we've been working on?

Great callout. I definitely wish we had a better way to test this query. My changes were based on [this thread](https://eppo-group.slack.com/archives/C060RP960P8/p1710871205479739?thread_ts=1710859262.362089&cid=C060RP960P8), which Bookbub tested and it eventually resolved their problem.

We should definitely prioritize setting up synthetic data in the demo workspace to test this. It would also be a great example to show that we are deeply familiar with the problem.

My thought here is that we know that the existing docs are wrong (from Bookbub), so it's good to resolve that issue ASAP

PR #373, opened 2024-06-03
Update Android documentation for UFC

aarsilv:
Update the Android documentation for UFC:
* New assignment method signatures (e.g., flag key first, default value)
* Describe the additional `extraLogging` and `metaData` assignment fields
* Remove section on handling `null`

I also expanded on some of the existing examples:
* Include initialization callback when initializing the SDK
* Include the subject key when logging an assignment


sameerank (approved):
:shipit:

PR #377, opened 2024-06-04
Added multiple comparisions explanation for dimensional SRM checks

tbuffington7:
Noticed that we don't explain how we address multiple comparisons in the dimensional SRM checks, so I added a brief explanation.


schmit (approved):
Nice!

PR #386, opened 2024-06-07
Add tip on choosing between COUNT DISTINCT and COUNT

giorgiomartini0:
We noticed that many users are choosing COUNT DISTINCT when COUNT would give similar results. This PR adds an info box about the higher warehouse costs and gives a more fleshed out example on how to choose between the two.


schmit (approved):
TIL about :::tip, this is a great one!

lukasgoetzweiss (approved):

PR #408, opened 2024-06-26
Add tip on disabling unused variants + reference for Bayesian

giorgiomartini0:
- Make it more explicit that we do multiple testing correction also across _variants_, and add tip on disabling unused variants
- Add details and reference on why Bayesian does not have MTC


emetelka (approved):

tbuffington7 (approved):
LGTM!

PR #412, opened 2024-07-01
chore(ruby): Add `Gemfile` instructions for using SDK

felipecsl:
Bundler with Gemfile is the most common way to consume gems in the ruby world

### DIFF
--- a/docs/sdks/server-sdks/ruby.md
+++ b/docs/sdks/server-sdks/ruby.md
@@ -16,6 +16,12 @@ Install the SDK with gem:
 gem install eppo-server-sdk
 ```
 
+or add to you `Gemfile`:
+
+```
+gem 'eppo-server-sdk, '~> 0.3.0'
+```
+
 ## 2. Initialize the SDK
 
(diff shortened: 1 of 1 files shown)


docs/sdks/server-sdks/ruby.md
```diff
@@ -16,6 +16,12 @@ Install the SDK with gem:
 gem install eppo-server-sdk
 ```
 
+or add to you `Gemfile`:
+
+```
+gem 'eppo-server-sdk, '~> 0.3.0'
```
aarsilv, line 22: I wonder if we should add a note to use the latest version as this will get out of date

aarsilv (approved):

docs/sdks/server-sdks/ruby.md
felipecsl: I think we should keep this in sync with the latest version as we publish new releases

felipecsl: could also remove the version specification, so Bundler will pick the latest, but it's good practice to pin the version

PR #421, opened 2024-07-17
Lg/anon

lukasgoetzweiss:


docs/guides/advanced-experimentation/anonymous-explainer.md
```diff
@@ -4,110 +4,90 @@ sidebar_position: 5
...
-FROM user_login_windows_collapsed
-ORDER BY user_id, ts_min;
+1. Filter to each Anonymous ID's first record during the experiment
+2. Check if the Anonymous ID was associated with more than one variant. If it was, remove it from downstream analysis
+3. For each Anonymous ID, determine the unique set of User IDs that was associated with it
+4. Join to both Anonymous ID and User ID metrics using the appropriate join column
+5. Aggregate metric events by Anonymous ID
 
-```
+Note that in step 3 we need to make sure each User ID is associated with only one Anonymous ID to avoid double counting post-authentication events (e.g., purchases). To ensure this, we associate each User ID with its *first* Anonymous ID during the course of the experiment.
```
tbuffington7, line 41: Should this be a numbered step above? Specifically, it sounds like there is a step before step 3 that is "For each User ID, identify the first record with a non-NULL Anonymous ID and exclude all other records."

I think we should clarify the ordering of the filtering too. For example, do we filter out duplicated rows for each User ID _before_  we filter out mixed anonymous IDs? I'm wondering if there's a case where this filtering step removes the record that would have made an Anonmous ID mixed.

docs/guides/advanced-experimentation/anonymous-explainer.md
```diff
@@ -4,110 +4,90 @@ sidebar_position: 5
...
+SELECT 
+      a.anonymous_id
+    , ta.user_id
+    , a.experiment
+    , a.variant
+    , a.timestamp
+   FROM assignments a 
+   LEFT JOIN timeboxed_associations ta
+     ON a.anonymous_id = ta.anonymous_id
+    AND ta.valid_from <= a.timestamp
```
tbuffington7, line 70: Should this have an `OR ta.valid_from IS NULL`? Thinking about the case where an anonymous ID is the first one for a given `user_id`, so the lag function would return `NULL`. Ditto for `valid_to`

docs/guides/advanced-experimentation/anonymous-explainer.md
```diff
@@ -4,110 +4,90 @@ sidebar_position: 5
...
 
-Within Eppo, first create an ‘anonymous’ entity. Feel free to give it whatever name makes the most sense for the organization. This entity will be used to organize both the Assignment SQL definition and the Fact SQL together.
+Second, it's undeniable that most users who have multiple anonymous IDs will never log in on every device. That is, there will inevitably be some Anonymous IDs that are indeed correlated but their relationship is never observed in event logs.
```
tbuffington7, line 82: This is true, but one might think "why not group the anonymous ids you know belong to the same user, even if we can't account for all of them?" It might be worth acknowledging the projected ID approach, even if it falls under the "Using your own existing attribution" approach

tbuffington7 (commented):
Looks good overall and I left a few comments. I'd like to spend more time understanding this better before approving, but if others approve, feel free to merge

docs/guides/advanced-experimentation/anonymous-explainer.md
lukasgoetzweiss: IMO it's more "how step 3 works" as opposed to a separate step

I don't think we dedupe records for User ID, just for anon ID. If user 1 has cookie A and cookie B, each with their own treatment, neither cookie A nor cookie B would be filtered out

docs/guides/advanced-experimentation/anonymous-explainer.md
lukasgoetzweiss: yep definitely

docs/guides/advanced-experimentation/anonymous-explainer.md
tbuffington7: I think I'm confused then. This line makes it sound like we ARE deduplicating records for user id?

> Note that in step 3 we need to make sure each User ID is associated with only one Anonymous ID 

I'm struggling to reconcile that line with this scenario:

> If user 1 has cookie A and cookie B, each with their own treatment, neither cookie A nor cookie B would be filtered out

Doesn't this mean that a purchase from User 1 could be double-counted across cookie A or B? I feel like I'm missing something subtle.

docs/guides/advanced-experimentation/anonymous-explainer.md
lukasgoetzweiss: added a small further justification. i agree we shouldn't shy away from the projected ID thing, but we do mention the BYO approach right below, so hopefully the reader will know they can keep using their own approach if they disagree with ours

docs/guides/advanced-experimentation/anonymous-explainer.md
lukasgoetzweiss: Oh yeah I think I'm getting tripped up. You're right, each cookie can only have multiple users, but each user can only have one cookie

So I see your point: do we check if a cookie is associated with multiple variants before we remove other userIDs or after

@jonathan-fulton probably can speak more to that part

tbuffington7: Hmm, I don't have a strong opinion, but it feels a little more intuitive to me to use a user's first non-mixed cookie. My understanding is that our approach follows from the assumption that these cookie-randomized experiments are generally independent across sessions. So, if a user has a non-mixed cookie that isn't impacted by a previous mixed cookie, it seems like we might as well include it.

jonathan-fulton: So after writing the SQL, it's a bit tricky to do the first non-mixed cookie.  Right now in the code we're just going to pick the first cookie a user is associated with.

jonathan-fulton (approved):

PR #422, opened 2024-07-18
Update background color

lukasgoetzweiss:


bertilhatt:
Looks good, changing the background to an egg-white, very light orange.

PR #424, opened 2024-07-18
Lg/whatiseppogif

lukasgoetzweiss:


felipecsl:
I just wanted to call out to the fact that the file is quite large (2mb), I'd recommend compressing it a bit if possible!

lukasgoetzweiss:
> I just wanted to call out to the fact that the file is quite large (2mb), I'd recommend compressing it a bit if possible!

PR #432, opened 2024-07-24
feat: [docs for ruby v3 ufc + bandits] (FF-2859)

leoromanovsky:
ruby v3 has been published with the UFC changes and bandits. I have modeled these docs on the python layout.

### DIFF
--- a/docs/sdks/server-sdks/ruby.md
+++ b/docs/sdks/server-sdks/ruby.md
@@ -8,7 +8,9 @@ Eppo's Ruby SDK can be used for both feature flagging and experiment assignment:
 - [GitHub repository](https://github.com/Eppo-exp/ruby-sdk)
 - [RubyGems gem](https://rubygems.org/gems/eppo-server-sdk/)
 
-## 1. Install the SDK
+## 1. Getting started
+
+### A. Install the SDK
 
 Install the SDK with gem:
 
@@ -19,33 +21,75 @@ gem install eppo-server-sdk
 or add to you `Gemfile`:
 
 ```
-gem 'eppo-server-sdk', '~> 0.3.0'
+gem 'eppo-server-sdk', '~> 3.0.0'
 ```
 
-## 2. Initialize the SDK
+### B. Initialize the SDK
 
-Initialize the SDK with a SDK key, which can be generated in the Eppo interface. Initialization the SDK when your application starts up to generate a singleton client instance, once per application lifecycle:
+To initialize the SDK, you will need an SDK key. You can generate one [in the flag interface](https://eppo.cloud/feature-flags/keys).
 
 ```ruby
 require 'eppo_client'
 
 config = EppoClient::Config.new('<YOUR_API_KEY>')
-client = EppoClient::init(config)
+EppoClient::init(config)
+client = EppoClient::Client.instance
 ```
 
-After initialization, the SDK begins polling Eppo’s API at regular intervals to retrieve the most recent experiment configurations such as variation values and traffic allocation. The SDK stores these configurations in memory so that assignments are effectively instant. For more information, see the [architecture overview](/sdks/overview) page.
+This generates a singleton client instance that can be reused throughout the application lifecycle.
 
-If you are using the SDK for experiment assignments, make sure to pass in an assignment logging callback (see [section](#define-an-assignment-logger-experiment-assignment-only) below).
+After initialization, the SDK begins polling Eppo's API at regular intervals to retrieve the most recent experiment configurations such as variation values and traffic allocation. The SDK stores these configurations in memory so that assignments are effectively instant. For more information, see the [architecture overview](/sdks/overview) page.
 
-:::info
+### C. Assign variations
 
-By default the Eppo client initialization is asynchronous to ensure no critical code paths are blocked. For more information on handling non-blocking initialization, see our [documentation here](/sdks/common-issues#3-not-handling-non-blocking-initialization).
+Assign users to flags or experiments using `get_<type>_assignment`, depending on the type of the flag.
+For example, for a String-valued flag, use `get_string_assignment`:
 
-:::
+```ruby
+require 'eppo_client'
 
-### Define an assignment logger (experiment assignment only)
+client = EppoClient::Client.instance
+variation = client.get_string_assignment(
+  '<FLAG-KEY>',
+  '<SUBJECT-KEY>',
+  {
+    # Optional map of subject metadata for targeting.
+  },
+  '<DEFAULT-VALUE>'
+)
+```
 
-If you are using the Eppo SDK for experiment assignment (i.e randomization), include a logger instance in the config that is passed to the `init` function on SDK initialization. The SDK invokes the `log_assignment` method in the instance to capture assignment data whenever a variation is assigned.
+The `get_string_assignment` function takes four required inputs to assign a variation:
+
+* `<FLAG-KEY>` is the key that you chose when creating a flag; you can find it on the [flag page](https://eppo.cloud/feature-flags). For the rest of this presentation, we'll use `"test-checkout"`. To follow along, we recommend that you create a test flag in your account, and split users between `"fast_checkout"` and `"standard_checkout"`.
+* `<SUBJECT-KEY>` is the value that identifies each entity in your experiment, typically `user_id`.
+* `<SUBJECT-ATTRIBUTES>` is a dictionary of metadata about the subject used for targeting. If you create targeting rules based on attributes, those attributes must be passed in on every assignment call. If no attributes are needed, pass in an empty dictionary.
+* `<DEFAULT-VALUE>` is the value that will be returned if no allocation matches the subject, if the flag is not enabled, if `get_string_assignment` is invoked before the SDK has finished initializing, or if the SDK was not able to retrieve the flag configuration.
+
+### Typed assignments
+
+Additional functions are available:
+
+```
+get_boolean_assignment(...)
+get_numeric_assignment(...)
+get_json_string_assignment(...)
+get_parsed_json_assignment(...)
+```
+
+Here's how this configuration looks in the [flag page](https://eppo.cloud/feature-flags):
+
+![Test checkout configuration](/img/feature-flagging/test-checkout-configuration.png)
+
+That's it: You can already start changing the feature flag on the page and see how it controls your code!
+
+However, if you want to run experiments, there's a little extra work to configure it properly.
+
+## 2. Assignment Logging for Experiment 
+
+If you are using the Eppo SDK for **experiment** assignment (i.e., randomization), we will need to know which entity, typically which user, passed through an entry point and was exposed to the experiment. For that, we need to log that information.
+
+Include a logger instance in the config that is passed to the `init` function on SDK initialization. The SDK invokes the `log_assignment` method in the instance to capture assignment data whenever a variation is assigned.
 
 The code below illustrates an example implementation of logging with Segment, but you could also use other event-tracking systems. The only requirement is that the SDK can call a `log_assignment` method. Here we override Eppo's `AssignmentLogger` class with a function named `log_assignment`, then instantiate a config using an instance of the custom logger class, and finally instantiate the client:
 
@@ -65,7 +109,17 @@ config = EppoClient::Config.new(
   '<YOUR_API_KEY>',
   assignment_logger: CustomAssignmentLogger.new
 )
-client = EppoClient::init(config)
+EppoClient::init(config)
+
+# Allow the configuration to be loaded from the CDN.
+sleep(1)
+
+variation = EppoClient::Client.instance.get_string_assignment(
+    flag_key,
+    subject_key,
+    {},
+    'default'
+)
 ```
 
 The SDK will invoke the `log_assignment` function with an `assignment` object that contains the following fields:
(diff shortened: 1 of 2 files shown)


docs/sdks/server-sdks/ruby.md
```diff
@@ -19,33 +21,75 @@ gem install eppo-server-sdk
...
 ```
-gem 'eppo-server-sdk', '~> 0.3.0'
+gem 'eppo-server-sdk', '~> 3.0.0'
 ```
 
-## 2. Initialize the SDK
+### B. Initialize the SDK
 
-Initialize the SDK with a SDK key, which can be generated in the Eppo interface. Initialization the SDK when your application starts up to generate a singleton client instance, once per application lifecycle:
+To initialize the SDK, you will need an SDK key. You can generate one [in the flag interface](https://eppo.cloud/feature-flags/keys).
```
aarsilv, line 29: minor, but would it be better to call it the "flags" interface (plural), or "configuration" as that's what they click to expose the SDK keys tab. OR we could just link to that!

image

docs/sdks/server-sdks/ruby.md
```diff
@@ -19,33 +21,75 @@ gem install eppo-server-sdk
...
+```ruby
+require 'eppo_client'
+
+client = EppoClient::Client.instance
+variation = client.get_string_assignment(
+  '<FLAG-KEY>',
+  '<SUBJECT-KEY>',
+  {
+    # Optional map of subject metadata for targeting.
+  },
```
aarsilv, line 57: I find `# Optional` confusing as later on you state we need to pass _something_, such as an empty dictionary. Maybe we should rephrase it to read like "Mapping of any subject metadata for targeting"

Note that Some of our other languages have syntactic sugar that let downstream developers skip it entirely, making the _parameter_ optional (as opposed to just its contents). Something we may want to do in rust/ruby land.

Lastly, in Ruby world should we say "hash" instead of "map" 🤔

docs/sdks/server-sdks/ruby.md
```diff
@@ -19,33 +21,75 @@ gem install eppo-server-sdk
...
+  '<DEFAULT-VALUE>'
+)
+```
+
+The `get_string_assignment` function takes four required inputs to assign a variation:
+
+* `<FLAG-KEY>` is the key that you chose when creating a flag; you can find it on the [flag page](https://eppo.cloud/feature-flags). For the rest of this presentation, we'll use `"test-checkout"`. To follow along, we recommend that you create a test flag in your account, and split users between `"fast_checkout"` and `"standard_checkout"`.
+* `<SUBJECT-KEY>` is the value that identifies each entity in your experiment, typically `user_id`.
+* `<SUBJECT-ATTRIBUTES>` is a dictionary of metadata about the subject used for targeting. If you create targeting rules based on attributes, those attributes must be passed in on every assignment call. If no attributes are needed, pass in an empty dictionary.
+* `<DEFAULT-VALUE>` is the value that will be returned if no allocation matches the subject, if the flag is not enabled, if `get_string_assignment` is invoked before the SDK has finished initializing, or if the SDK was not able to retrieve the flag configuration.
```
aarsilv, line 67: My impression from the slack thread wa that "no allocation matches the subject" is technically an error case given we have default allocations? So perhaps we should leave that out and roll it into a general "or the SDK encountered an error determining the assignment"

docs/sdks/server-sdks/ruby.md
```diff
@@ -65,7 +109,17 @@ config = EppoClient::Config.new(
   '<YOUR_API_KEY>',
   assignment_logger: CustomAssignmentLogger.new
 )
-client = EppoClient::init(config)
+EppoClient::init(config)
+
+# Allow the configuration to be loaded from the CDN.
+sleep(1)
```
aarsilv, line 115: 🤨 no awaits or callbacks in ruby? or can we make a synchronous version?

Don't love seeing `sleep()` in our example 😞

docs/sdks/server-sdks/ruby.md
```diff
@@ -84,71 +138,152 @@ The SDK will invoke the `log_assignment` function with an `assignment` object th
...
-client = EppoClient::Client.instance
-variation = client.get_string_assignment(
-  '<SUBJECT-KEY>',
-  '<FLAG-KEY>',
-  {
-    # Optional map of subject metadata for targeting.
-  }
-)
-```
+At initialization, the SDK polls Eppo's API to retrieve the most recent experiment configuration. The SDK stores that configuration in memory. This is why assignments are effectively instant, as you can see yourself by profiling the code above.
```
aarsilv, line 147: technically, it polls our CDN, hydrated from our API. I'ts probably worth calling out that distinction.

docs/sdks/server-sdks/ruby.md
```diff
@@ -84,71 +138,152 @@ The SDK will invoke the `log_assignment` function with an `assignment` object th
...
-```
+### B. Automatically Updating the SDK Configuration
+
+After initialization, the SDK continues polling Eppo's API at 30-second intervals. This retrieves the most recent flag and experiment configurations such as variation values, targeting rules, and traffic allocation. This happens independently of assignment calls.
 
-### Handling `nil`
+:::note
 
-We recommend always handling the `nil` case in your code. Here are some examples of when the SDK returns `nil`:
+Changes made to experiments on Eppo's web interface are almost instantly propagated through our Content-delivery network (CDN) Fastly. Because of the refresh rate, it may take up to 30 seconds (± 5 seconds fuzzing) for those to be reflected by the SDK assignments.
```
aarsilv, line 163: consider the term "jitter" instead of "fuzzing"

docs/sdks/server-sdks/ruby.md
```diff
@@ -84,71 +138,152 @@ The SDK will invoke the `log_assignment` function with an `assignment` object th
...
+### A. Add a bandit action logger to the assignment logger
+
+In order for the bandit to learn an optimized policy, we need to capture and log the bandit actions.
+This requires adding a bandit action logging callback to the AssignmentLogger class
+```ruby
+class MyLogger < EppoClient::AssignmentLogger
+    def log_assignment(assignment):
+        ...
+
+    def log_bandit_action(self, bandit_action):
```
aarsilv, line 189: why the `self` argument?

In other SDKs and their documentation, we call the parameters `assignmentEvent` and `banditEvent`, as they contain more than just the assignment and action respectively.

docs/sdks/server-sdks/ruby.md
```diff
@@ -84,71 +138,152 @@ The SDK will invoke the `log_assignment` function with an `assignment` object th
...
+| `flagKey` (String)                                | The key of the feature flag corresponding to the bandit                                                                                           | "bandit-test-allocation-4"          |
+| `banditKey` (String)                                       | The key (unique identifier) of the bandit                                                                       | "ad-bandit-1"                       |
+| `subject` (String)                                   | An identifier of the subject or user assigned to the experiment variation                                       | "ed6f85019080"                      |
+| `action` (String)                                    | The action assigned by the bandit                                                                               | "promo-20%-off"                     |
+| `subjectNumericAttributes` (Hash{String => Float})     | Metadata about numeric attributes of the subject. Hash of the name of attributes their numeric values            | `{"age": 30}`    |
```
aarsilv, line 202: I feel like we're missing a word here. I think we want something like "hash of the name of the attributes **_to_** their numeric values" or "hash of the name of the attributes **_and_** their numeric values"

docs/sdks/server-sdks/ruby.md
```diff
@@ -84,71 +138,152 @@ The SDK will invoke the `log_assignment` function with an `assignment` object th
...
+| `actionNumericAttributes` (Hash{String => Float})      | Metadata about numeric attributes of the assigned action. Hash of the name of attributes their numeric values    | `{"discount": 0.1}`   |
+| `actionCategoricalAttributes` (Hash{String => String})  | Metadata about non-numeric attributes of the assigned action. Hash of the name of attributes their string values | `{"promoTextColor": "white"}` |
+| `actionProbability` (Float)                         | The weight between 0 and 1 the bandit valued the assigned action                                                | 0.25                                |
+| `modelVersion` (String)                              | Unique identifier for the version (iteration) of the bandit parameters used to determine the action probability | "v123"                       |
+
+### B. Querying the bandit for an action
+
+To query the bandit for an action, you can use the `get_bandit_action` function. This function takes the following parameters:
+- `flag_key` (str): The key of the feature flag corresponding to the bandit
```
aarsilv, line 212: Rubify to `(String)`?

docs/sdks/server-sdks/ruby.md
```diff
@@ -84,71 +138,152 @@ The SDK will invoke the `log_assignment` function with an `assignment` object th
...
+        # implement me
+```
+
+We automatically log the following data:
+
+| Field                                                | Description                                                                                                     | Example                             |
+|------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------|-------------------------------------|
+| `timestamp` (Date)                                   | The time when the action is taken in UTC  | 2024-03-22T14:26:55.000Z            |
+| `flagKey` (String)                                | The key of the feature flag corresponding to the bandit                                                                                           | "bandit-test-allocation-4"          |
+| `banditKey` (String)                                       | The key (unique identifier) of the bandit                                                                       | "ad-bandit-1"                       |
```
aarsilv, line 199: does our Ruby event use cammel case property names? Should we make it snake for Ruby land? If we don't have a translation layer, given the difference in language conventions, I wonder if event case is some setting we pass through to rust.

docs/sdks/server-sdks/ruby.md
```diff
@@ -84,71 +138,152 @@ The SDK will invoke the `log_assignment` function with an `assignment` object th
...
+| `actionCategoricalAttributes` (Hash{String => String})  | Metadata about non-numeric attributes of the assigned action. Hash of the name of attributes their string values | `{"promoTextColor": "white"}` |
+| `actionProbability` (Float)                         | The weight between 0 and 1 the bandit valued the assigned action                                                | 0.25                                |
+| `modelVersion` (String)                              | Unique identifier for the version (iteration) of the bandit parameters used to determine the action probability | "v123"                       |
+
+### B. Querying the bandit for an action
+
+To query the bandit for an action, you can use the `get_bandit_action` function. This function takes the following parameters:
+- `flag_key` (str): The key of the feature flag corresponding to the bandit
+- `subject_key` (str): The key of the subject or user assigned to the experiment variation
+- `subject_attributes` (Attributes): The context of the subject
```
aarsilv, line 214: In other SDKs we allow either freeform attributes ([key] => bool | number | string) or contextual attributes (numeric_attributes => [key] => number, categorical_attributes => [key] => bool | number | string).

Assuming we do here (and if not, we should adjust) we should call this out in the documentation. It may also help to right after this explain the difference, as we do in our JavaScript documentation ([link](https://docs.geteppo.com/sdks/server-sdks/node/#querying-the-bandit-for-an-action)).

docs/sdks/server-sdks/ruby.md
```diff
@@ -84,71 +138,152 @@ The SDK will invoke the `log_assignment` function with an `assignment` object th
...
+| `actionProbability` (Float)                         | The weight between 0 and 1 the bandit valued the assigned action                                                | 0.25                                |
+| `modelVersion` (String)                              | Unique identifier for the version (iteration) of the bandit parameters used to determine the action probability | "v123"                       |
+
+### B. Querying the bandit for an action
+
+To query the bandit for an action, you can use the `get_bandit_action` function. This function takes the following parameters:
+- `flag_key` (str): The key of the feature flag corresponding to the bandit
+- `subject_key` (str): The key of the subject or user assigned to the experiment variation
+- `subject_attributes` (Attributes): The context of the subject 
+- `actions` (Hash{String => Attributes}): A hash that maps available actions to their attributes
```
aarsilv, line 215: Same for actions, except we also allow just a set of Strings too if we don't have any action context.

docs/sdks/server-sdks/ruby.md
```diff
@@ -84,71 +138,152 @@ The SDK will invoke the `log_assignment` function with an `assignment` object th
...
+Note that action contexts can contain two kinds of information:
+- Action-specific context: e.g., the image aspect ratio of the image corresponding to this action
+- User-action interaction context: e.g., there could be a "brand-affinity" model that computes brand affinities of users to brands, and scores of this model can be added to the action context to provide additional context for the bandit.
+
+#### Result
+
+The `bandit_result` is an instance of `BanditResult`, which has two fields:
+
+- `variation` (String): The variation that was assigned to the subject
+- `action` (Optional[String]): The action that was assigned to the subject
```
aarsilv, line 276: is `Optional[String]` Ruby-speek?

docs/sdks/server-sdks/ruby.md
```diff
@@ -84,71 +138,152 @@ The SDK will invoke the `log_assignment` function with an `assignment` object th
...
+
+The variation returns the feature flag variation; this can be the bandit itself, or the "status quo" variation if the user is not assigned to the bandit.
+If we are unable to generate a variation, for example when the flag is turned off, then the `default` variation is returned. 
+In both of those cases, the `action` is `nil`, and you should use the status-quo algorithm to select an action.
```
aarsilv, line 280: this may not make sense because one of the above cases is the bandit itself, in which case `action` won't be `nil`.

Perhaps reword to "In both of those cases, when `action` is `nil`, you should use the status-quo algorithm to select an action.

docs/sdks/server-sdks/ruby.md
```diff
@@ -84,71 +138,152 @@ The SDK will invoke the `log_assignment` function with an `assignment` object th
...
+
+#### Result
+
+The `bandit_result` is an instance of `BanditResult`, which has two fields:
+
+- `variation` (String): The variation that was assigned to the subject
+- `action` (Optional[String]): The action that was assigned to the subject
+
+The variation returns the feature flag variation; this can be the bandit itself, or the "status quo" variation if the user is not assigned to the bandit.
+If we are unable to generate a variation, for example when the flag is turned off, then the `default` variation is returned.
```
aarsilv, line 279: There is another case, where the bandit encounters an error selecting an action (or there are no actions provided) in which case it will return the bandit variation, but a `nil` action. I wonder if that's worth explicitly calling out--I think it is.

aarsilv (approved):
Nice work! 🎉  Approving with a BOATLOAD 🛥️ of comments you are welcome to act on or ignore as you see fit!

docs/sdks/server-sdks/ruby.md
felipecsl: agreed, ruby is single threaded by default though and doesnt have async primitives like async/await. we could have a callback instead, but that would require changing the interface

docs/sdks/server-sdks/ruby.md
felipecsl: maybe a python-ism?

leoromanovsky: Thanks for noticing - sorry I copied this from python and missed this.

PR #434, opened 2024-07-30
Android SDK installation instructions point to 3.+

giorgiomartini0:
Instead of specific version. All changes on `3.x.x` will be non-breaking.

### DIFF
--- a/docs/sdks/client-sdks/android.md
+++ b/docs/sdks/client-sdks/android.md
@@ -13,7 +13,7 @@ Eppo's open source Android SDK can be used for both feature flagging and experim
 You can install the SDK using Gradle by adding to your `build.gradle` file:
 
 ```groovy
-implementation 'cloud.eppo:android-sdk:3.2.0'
+implementation 'cloud.eppo:android-sdk:3.+'
 ```
 
 ## 2. Initialize the SDK


docs/sdks/client-sdks/android.md
```diff
@@ -13,7 +13,7 @@ Eppo's open source Android SDK can be used for both feature flagging and experim
 You can install the SDK using Gradle by adding to your `build.gradle` file:
 
 ```groovy
-implementation 'cloud.eppo:android-sdk:3.2.0'
+implementation 'cloud.eppo:android-sdk:3.+'
```
felipecsl, line 16: FYI it's not recommended to use `+` syntax on Gradle as it leads to non-deterministic behavior with unexpected changing library versions. I know this is a different convention than what folks usually do for NPM/JS ecossystem, for example, but on JVM world people are a lot more conscious of this stuff. Just wanted to call that out

felipecsl: Just a data point:
image
https://developer.vonage.com/en/blog/the-right-way-of-using-gradle-dynamic-dependencies

giorgiomartini0: Good callout - I didn't know. Reverting to point explicitly to the latest version.

felipecsl: thank you 🙏🏽

PR #439, opened 2024-08-07
Eval details edits

hhargreaveseppo:


docs/sdks/sdk-features/debugging-flag-assignment.md
```diff
@@ -54,6 +54,34 @@ getStringAssignment(flagKey, subjectKey, subjectAttributes, defaultValue);
...
+```
+
+and a Numeric flag will have:
+
+```typescript
+variationKey: '3',
+variationValue: 3
+```
+
+An exception exists for JSON flags where the `variationKey` will be the variation name given in the flag:
```
schmit, line 75: Note: it's not exactly the name (lower cased and dashes added), so "My Variation" becomes "my-variation" -- and it doesn't change when the name gets updated. Not sure how much of this detail you want to add

hhargreaveseppo: ah good catch -- i'll add those, i think it's important for customers to know

docs/sdks/sdk-features/debugging-flag-assignment.md
```diff
@@ -82,6 +82,8 @@ variationValue: {
 }
 ``` 
 
+Additionally, the string generated for the `varationKey` on a JSON flag will be converted to lower case and spaces will be replaced with `-`'s. `Model 1` as a variation name on the flag will become `model-1` in the evaluation details and the assignment logger `variationKey`. This value is static and will not update if the variation name in the JSON flag is updated. If `Model 1` is updated to `ChatGPT Model`, `model-1` will continue to be logged.
+
```
hhargreaveseppo, line 86: How's this @schmit?

PR #447, opened 2024-08-30
update okta page with info block about IdP-initiated logins

btthomas:
Some customers are assuming that we support IdP-initiated logins via an Okta tile. This warning plus copy change should help.

Screenshot 2024-08-30 at 10 18 01 AM

I would prefer `warning`, but that renders a `danger`, so I used `info` instead.

### DIFF
--- a/docs/administration/okta.md
+++ b/docs/administration/okta.md
@@ -3,10 +3,6 @@ sidebar_position: 4
 ---
 # Authenticating with Okta
 
-:::info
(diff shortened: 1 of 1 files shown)


emetelka (commented):
Can we move this further down the page? I'd prefer not to have the first thing someone sees is information on a feature we don't have

btthomas:
I moved the info box to after the steps.

lukasgoetzweiss (approved):

PR #448, opened 2024-08-30
Add details for guardrail metrics

emetelka:
Also made an update to metric windows to be more precise for @tbuffington7


docs/data-management/metrics/index.md
```diff
@@ -21,9 +21,14 @@ $$
 
 [Funnels](/data-management/metrics/funnel-metric) are another powerful metric type in Eppo. They allow you to track and analyze multi-step processes or user flows. By defining the sequence of events in a funnel, you can identify bottlenecks, drop-off points, and conversion rates at each step. This facilitates a granular analysis of user behavior and helps optimize the customer journey.
 
-## Guardrail metrics
 
-Furthermore, you can set metrics as [guardrail metrics](/data-management/organizing-metrics/guardrails), which means they are automatically added to every experiment. This feature ensures that specific metrics are consistently tracked across experiments, providing a standardized measurement framework. By setting metrics as guard rails, you can maintain a unified approach to experimentation and easily compare results across different tests.
+## Guardrail cutoffs
+
+You can set metrics to have a [guardrail cutoff](/data-management/organizing-metrics/guardrails), which means they will appear with a downside threshold. When the results for the metric breach the threshold, the metric will display a warning state.
```
schmit, line 27: This talks about the "what", but doesn't talk about the "why" -- I think we should add something about the motivation regarding non-inferiority tests

docs/data-management/organizing-metrics/guardrails.md
```diff
@@ -2,18 +2,29 @@
 sidebar_position: 6
 ---
 
-# Guardrail metrics
+# Guardrail cutoffs
 
-A guardrail metric refers to a specific type of metric in Eppo that is set as a standard measurement for every experiment.
-By configuring a metric as a guardrail, it is automatically included in the analysis of all applicable experiments, ensuring consistent tracking and comparison of results.
-This approach enables you to establish a uniform measurement framework, allowing you to monitor and evaluate key performance indicators consistently across various tests, providing valuable insights and facilitating data-driven decision-making processes.
+A guardrail cutoff refers to a downside threshold that can be set on a metric. When the results for the metric breach this threshold, the metric will display a yellow warning state.
```
schmit, line 7: I think the wording here is a bit awkward and confusing. Instead I would reword to something like

The guardrail cutoff metric is yellow until we are fairly certain that the downside risk on this metric is sufficiently small.

docs/data-management/organizing-metrics/guardrails.md
```diff
@@ -2,18 +2,29 @@
...
+
+Use a guardrail cutoff where a neutral result is okay but a downside risk should be avoided. They also enable non-inferiority testing when a metric with a guardrail is set as primary.
```
schmit, line 9:
```suggestion
Guardrail cut-off metrics are primarily useful to track metrics that an experiment is not supposed to positively impact, but you want to ensure that there is no negative effect either. A common use case is page load times: few experiments explicitly want to improve page load times, but likely most experiments want to avoid slowing down page load times. Guardrail cut-offs also enable non-inferiority testing when a metric with a guardrail is set as primary.
```

docs/data-management/organizing-metrics/guardrails.md
```diff
@@ -2,18 +2,29 @@
...
+## Guardrail metrics in experiments
+
+![Guardrail metrics in an experiment](/img/data-management/metrics/guardrail-experiment.png)
+
+A metric with a guardrail cutoff will appear with a `Guardrail label` and a grey boundary indicating the cutoff zone.
+
+When a metric does not breach the cutoff threshold and is not statistically significant positive or negative, it will appear as grey with a check that the cutoff was not breached.
 
-2. Click on the No/Yes dropdown in the **Is Guardrail** column
+When a metric breaches the cutoff threshold and is not statistically significant negative, it will appear as yellow with a warning that the cutoff was breached. In this case, you should consider not rolling out the variant given that the Guardrail metric was not above the pre-determined threshold.
```
schmit, line 28: I think we should define what "breaches" mean here a bit more carefully. In particular, it's about the lower bound of the confidence interval

docs/data-management/organizing-metrics/guardrails.md
```diff
@@ -2,18 +2,29 @@
...
 
-3. Select **Yes** from the dropdown menu
+When a metric breaches the cutoff threshold and not statistically significant negative, it will appear as red just as any other statistically significant negative metric would.
```
schmit, line 30:
```suggestion
When a metric breaches the cutoff threshold and is statistically significant negative, it will appear as red just as any other statistically significant negative metric would.
```

PR #459, opened 2024-09-10
Replacing link to Google Docs to a table

bertilhatt:
A table would be easier to understand.


bertilhatt:
Continued https://github.com/Eppo-exp/eppo-docs/pull/461

PR #466, opened 2024-09-12
Mention Guardrail feature on the Non inferiority test page

bertilhatt:
We probably should mention that we support non-inferiority tests on a page that explains how to do it yourself.

Alternatively, we could remove that page, but I think it’s reasonable to keep it for a bit in the event users have kept links to this page to point them at the page about the feature.

### DIFF
--- a/docs/guides/advanced-experimentation/non-inferiority-tests.md
+++ b/docs/guides/advanced-experimentation/non-inferiority-tests.md
@@ -1,5 +1,9 @@
 # Running Non-inferiority Tests
 
+:::info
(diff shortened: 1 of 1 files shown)


emetelka:
We should update this page substantively beyond saying that we now have guardrails. These guides should be long lived and approachable to someone who is not familiar with Eppo.

bertilhatt:
I am happy to do it, but I am not sure which direction you want to take it.
What changes do you want to suggest? More approachable than now? Not tied
to how Eppo presents results?

lukasgoetzweiss (approved):
We should delete this whole page soon, but merging this in for now as a temporary stop gap

PR #471, opened 2024-09-25
adding SDK clarification

hhargreaveseppo:


docs/feature-flagging/concepts/mutual_exclusion.md
```diff
@@ -57,6 +57,64 @@ Here you are able to define the Experiment name, the Variations to include in th
...
+```
+
+will return:
+
+```javascript
+[
+  { parameterName: 'sticky_banner', variationValue: 'false' },
+  { parameterName: 'banner_cta', variationValue: 'Sign in for updates on shipping' },
+  { parameterName: 'shoprunner', variationValue: 'true' },
+  { parameterName: 'promo', variationValue: '$20 off $100' }
```
schmit, line 82: @aarsilv and I want to change this return type because it is extremely unwieldy for users, to instead return

```
{'sticky_banner': 'false', 'banner_cta': 'Sign in for updates on shipping', ...}
```

lukasgoetzweiss (approved):

docs/feature-flagging/concepts/mutual_exclusion.md
hhargreaveseppo: sweet! good to know and I agree. I'll keep this as is for now so we can highlight to users the difference between a JSON flag and a layer.

schmit: we basically do not want anyone adopting this approach as it would make it impossible for us to make this change

PR #478, opened 2024-09-30
Add sequential hybrid tost explanation

tbuffington7:
Added an explanation about using sequential hybrid as two one-sided tests.


docs/statistics/confidence-intervals/analysis-methods.md
```diff
@@ -335,6 +335,14 @@ The sequential hybrid option aims to take the best of both worlds: continuous mo
...
+Significant degradations due to poor user experiences also often have large effect sizes that offset the loss of power from the sequential methodology. In these cases, the point estimates for the lift are also of less interest compared to experiments with "winning" variants.
+
+Conversely, for detecting improvements, it is often helpful to have additional power and to have more reliable estimates of the treatment effect, which are both advantages of the fixed-sample approach. As a result, a sensible approach is to use sequential hybrid's sequential test for early detection of poorly performing variants 
+and its fixed-sample approach for detecting improvements. This approach is effectively two one-sided tests: a sequential test with a significance level $\frac{\alpha}{4}$ is performed continuously on the degradation tail and a fixed-sample test with a significance level $\frac{\alpha}{4}$ is performed on the experiment's end date on the improvement tail.
```
schmit, line 343: ```This approach is effectively two one-sided tests: a sequential test with a significance level $\frac{\alpha}{4}$ is performed continuously on the degradation tail and a fixed-sample test with a significance level $\frac{\alpha}{4}$ is performed on the experiment's end date on the improvement tail.```

I think this is pretty confusing, in particular, I imagine people will wonder: so should I increase my alpha? How should I think of type I error rates? I think you should either explain more or cut it out

schmit (approved):
Nice! added one comment

docs/statistics/confidence-intervals/analysis-methods.md
tbuffington7: Great callout @schmit. I decided to take the path of explaining rather than omitting because if I don't explain it here, I'll need to explain it later 😄 

When you get a chance, could you take another look after this [this commit](https://github.com/Eppo-exp/eppo-docs/commit/b23fa1f3f35db434a1b43245065618990b95c744)?

docs/statistics/confidence-intervals/analysis-methods.md
```diff
@@ -335,6 +335,19 @@ The sequential hybrid option aims to take the best of both worlds: continuous mo
...
+
+Conversely, for detecting improvements, it is often helpful to have additional power and to have more reliable estimates of the treatment effect, which are both advantages of the fixed-sample approach. As a result, a sensible approach is to use sequential hybrid's sequential test for early detection of poorly performing variants 
+and its fixed-sample approach for detecting improvements. 
+
+This approach is effectively two one-sided tests: a sequential test with a significance level $\frac{\alpha}{4}$ is performed continuously on the degradation tail and a fixed-sample test with a significance level $\frac{\alpha}{4}$ is performed on the experiment's end date on the improvement tail. To understand where the $\frac{\alpha}{4}$ comes from, first recognize that
+the core idea of the sequential hybrid methodology is that we allocate half of the "alpha budget" to the sequential test and half to the fixed sample test. This means that the two-tailed sequential test has a significance level $\frac{\alpha}{2}$ and half of _that_ alpha is allocated to each tail, leaving $\frac{\alpha}{4}$ for each tail. If we stop the test early only for degradations,
```
schmit, line 346:
```suggestion
the core idea of the sequential hybrid methodology is that we allocate half of the "$\alpha$ budget" to the sequential test and half to the fixed sample test. This means that the two-tailed sequential test has a significance level $\frac{\alpha}{2}$ and half of _that_ $\alpha$ is allocated to each tail, leaving $\frac{\alpha}{4}$ for each tail. If we stop the test early only for degradations, 
```

schmit (approved):

PR #487, opened 2024-10-10
docs for the Eppo Contentful marketplace app

greghuels:


hhargreaveseppo:
This looks great! Two follow up questions:
1. **How would targeting work?** If a user wanted to target a Contentful test to only users with a specific cookie value, could they log into Eppo, update the targeting condition on experiment allocation they just created and then make sure that the user context is passed in where ever getStringAssignment() is called? Not even sure if the user would be able to grok access to where the assignment is made.
2. **Are bandits supported in this iteration?** I know one of the original requests for the Contentful integration was to support Bandit testing for a particular customer, is that in scope for this version? If not, will we expect it in a later version? This question is not as pressing, I'm not aware of any other customers that are waiting on Bandits to be available in Contentful.

greghuels:
> This looks great! Two follow up questions:
> 
> 1. **How would targeting work?** If a user wanted to target a Contentful test to only users with a specific cookie value, could they log into Eppo, update the targeting condition on experiment allocation they just created and then make sure that the user context is passed in where ever getStringAssignment() is called? Not even sure if the user would be able to grok access to where the assignment is made.
> 2. **Are bandits supported in this iteration?** I know one of the original requests for the Contentful integration was to support Bandit testing for a particular customer, is that in scope for this version? If not, will we expect it in a later version? This question is not as pressing, I'm not aware of any other customers that are waiting on Bandits to be available in Contentful.

Great questions. For targeting, a user could indeed make changes in the Eppo UI to whatever rules they want. And we do still allow developers to pass in a user subject key and context.

To your second question, bandits are not supported in this iteration. This is a basic "version 1" that we wanted to get out the door. However, we could add support for bandits in a subsequent iteration if that's something that gets prioritized.

docs/guides/marketing/integrating-with-contentful.md
```diff
@@ -6,115 +6,60 @@ sidebar_position: 6
...
+Using a headless CMS like Contentful, and a lightweight, robust feature flagging SDK, like Eppo, allows the best of both worlds: an easy way for marketing teams to run tests without sacrificing performance. In this tutorial, we will walk through how to set up Eppo’s Node SDK so Marketing teams can run their own experiments by updating [Content](https://www.contentful.com/help/content-model-and-content-type/) defined in Contentful and feature flags in Eppo.
 
+## Installing Eppo's Contentful marketplace app
 
-## Setting up Contentful
+Eppo now supports seamless integration with Contentful via the Eppo Contentful marketplace app. This app allows you to create experiements directly within a Contentful blog post.
 
-For this example, imagine we are running an experiment on the Eppo homepage. There are a few places where we might want to experiment on this page:
+To get started, search for "Eppo" on Contentful's marketplace. To navigate to the marketplace, click on **Apps** from Contentful's navigation menu, then click **Marketplace**.
```
felipecsl, line 15: we could probably include a link to the actual app page here once it's live. just leaving as a reminder for our future selves

felipecsl (approved):
Awesome work, looks great. I'd say let's wait to merge until the Contentful app is live to avoid confusing users in the meantime

docs/guides/marketing/integrating-with-contentful.md
```diff
@@ -6,115 +6,66 @@ sidebar_position: 6
...
-2. Create a new variant in Eppo with the `entry_id` from Contentful UI.
-3. Create an allocation to test the new content in production (see screenshot below).
-4. Load page to QA new content and [add screenshots to Eppo](/experiment-analysis/configuration/#adding-screenshots) for reference.
-5. [Create an experiment allocation](/feature-flagging/concepts/experiment-assignment/) and launch the experiment.
-6. Analyze experiment and make rollout decisions like any other Eppo experiment.
-
-:::tip
-This can all be done in the Contentful and Eppo UIs with no additional engineering support.
+:::danger
+The experiment is not running until the linked fetaure flag is enabled in the Eppo UI. Don't forget to enable the feature flag.
```
felipecsl, line 62: typo: `fetaure`

PR #497, opened 2024-10-17
Color & spacing polish

hang-design:
- Added brand colors as  --ifm color variables
- Swapped them to the left nav & toc color background. Made bg use gray-300 because it needs a bit more contrast
- Fixed padding where the hover was touching the right side of sidebar
- Added left nav border back to create separation between nav and content in dark mode
- Changed footer background color to match brand color
- Added a bit more space to horizontal padding for main content area


lukasgoetzweiss (approved):
looks great!

PR #501, opened 2024-10-22
Data pipeline: add clustered analysis diagram and description

jonathan-fulton:
Add clustered analysis diagram and description.


htrajan (approved):

lisaah (approved):
🎉  

Just something (preexisting) I noticed, not sure worth updating, is that we also use funnel data in cuped but that step isn't in our docs.

jonathan-fulton:
Here's the preview screenshot.

image

jonathan-fulton:
> 🎉
> 
> Just something (preexisting) I noticed, not sure worth updating, is that we also use funnel data in cuped but that step isn't in our docs.

@lisaah yea, I didn't know that when I made the original diagram.  Maybe I'll update that separately b/c it'll require me to re-order the pipeline branches to keep it clean.

PR #509, opened 2024-10-30
chore:update central docs for dotnet

typotter:


docs/sdks/server-sdks/dotnet.md
```diff
@@ -3,39 +3,65 @@ import TabItem from '@theme/TabItem';
 
 # .NET
 
-Eppo's open source .NET SDK can be used for both feature flagging and experiment assignment:
+Eppo's open source .NET SDK can be used for feature flagging and experiment assignment and Multi-Armed Contextual Bandits:
```
aarsilv, line 6: ❤️ 

Any particular reason you capitalized it?

docs/sdks/server-sdks/dotnet.md
```diff
@@ -3,39 +3,65 @@ import TabItem from '@theme/TabItem';
...
 
 - [GitHub repository](https://github.com/Eppo-exp/dot-net-server-sdk)
 - [Package](https://www.nuget.org/packages/Eppo.Sdk)
+- [Nuget Package](https://www.nuget.org/packages/Eppo.Sdk)
```
aarsilv, line 10: 🔥

docs/sdks/server-sdks/dotnet.md
```diff
@@ -3,39 +3,65 @@ import TabItem from '@theme/TabItem';
...
-## 1. Install the SDK
+
+## Getting Started
+
+### Installation
 
 In your .NET application, add the Eppo.Sdk Package from Nuget.
 
-```
+```shell
```
aarsilv, line 19: 🙏 I appreciate the syntax highlighting hints for GitHub code blocks

docs/sdks/server-sdks/dotnet.md
```diff
@@ -3,39 +3,65 @@ import TabItem from '@theme/TabItem';
...
-Initialize the SDK with a SDK key, which can be generated in the Eppo interface. Initialization should happen when your application starts up to generate a singleton client instance, once per application lifecycle:
+Begin by initializing a singleton instance of Eppo's client with an SDK key from the [Eppo interface](https://eppo.cloud/feature-flags/keys). Once initialized, the client can be used to make assignments anywhere in your app. Initialization should happen when your application starts up to generate a singleton client instance, once per application lifecycle:
 
-Then use that to create the EppoClient instance.
+
+#### Initialize once
 
 ```csharp
-var eppoClientConfig = new EppoClientConfig(apiToken, new AssignmentLogger());
+var eppoClientConfig = new EppoClientConfig("<SDK_KEY>", new AssignmentLogger()); // Don't worry about the `AssignmentLogger` just yet. More on it below.
```
aarsilv, line 31: Could we just pass `null` for the assignment logger here for now? I reckon in the (small minority) use case they are only using feature flags for feature control and not experimentation, they wouldn't need an assignment logger, so it would technically be valid. Then the last paragraph of this section leads into "If you are using the SDK for experiment assignments..."

docs/sdks/server-sdks/dotnet.md
```diff
@@ -3,39 +3,65 @@ import TabItem from '@theme/TabItem';
...
+  "<SDK_KEY>",
+  new AssignmentLogger());
+```
 
-If you are using the Eppo SDK for experiment assignment (i.e randomization), pass in a callback logging function on SDK initialization. The SDK invokes the callback to capture assignment data whenever a variation is assigned.
+The code below illustrates an example implementation of a logging callback to the console and other event platforms such as Segment.  You could also use your own logging system, the only requirement is that the SDK receives a `LogAssignment(AssignmentLogData assignmentLogData)` function. The `AssignmentLogData` class implements `ISerializable` so most systems should be able to easily log the value.
 
-The code below illustrates an example implementation of a logging callback using Segment. You could also use your own logging system, the only requirement is that the SDK receives a `LogAssignment` function. Here we define an implementation of the Eppo `IAssignmentLogger` interface containing a single function named `LogAssignment`:
+<Tabs>
+<TabItem value="console" label="Console">
```
aarsilv, line 64: 🔥

docs/sdks/server-sdks/dotnet.md
```diff
@@ -49,60 +75,262 @@ internal class AssignmentLogger : IAssignmentLogger
...
-The `GetStringAssignment` function takes two required and one optional input to assign a variation:
+Each function has the same signature (except for the type of `defaultValue`) and returns the type in the function name. For booleans use `getBooleanAssignment`, which has the following signature:
+
+```cs
+public bool GetBooleanAssignment(
+    string flagKey, 
+    string subjectKey, 
+    Dictionary<string, object> subjectAttributes, 
+    bool defaultValue
+)
```
aarsilv, line 131: should there be semicolon after this closing parenthesis?

docs/sdks/server-sdks/dotnet.md
```diff
@@ -49,60 +75,262 @@ internal class AssignmentLogger : IAssignmentLogger
...
 
-Starting in version `v2.0.0` typed functions are available:
+The SDK will invoke the `LogAssignment` function with an `event` object that contains the following fields:
+
+| Field                                     | Description                                                                                                              | Example                             |
+| ----------------------------------------- | ------------------------------------------------------------------------------------------------------------------------ | ----------------------------------- |
+| `Experiment` (string)                     | An Eppo experiment key                                                                                                   | "recommendation-algo-allocation-17" |
+| `Subject` (string)                        | An identifier of the subject or user assigned to the experiment variation                                                | UUID                                |
```
aarsilv, line 156: I wonder if we should have an example instead of just `UUID` (such as a nano ID `"V1StGXR8_Z5jdHi6B-myT"`)

Update: we do this later in the bandit assignment schema `"ed6f85019080"`

docs/sdks/server-sdks/dotnet.md
```diff
@@ -49,60 +75,262 @@ internal class AssignmentLogger : IAssignmentLogger
...
+| `Variation` (string)                      | The experiment variation the subject was assigned to                                                                     | "control"                           |
+| `Timestamp` (DateTime)                    | The time when the subject was assigned to the variation                                                                  | 2021-06-22T17:35:12.000Z            |
+| `SubjectAttributes` (Map<String, object>) | A free-form map of metadata about the subject. These attributes are only logged if passed to the SDK assignment function | `Map.of("device","iOS")`            |
+| `FeatureFlag` (string)                    | An Eppo feature flag key                                                                                                 | "recommendation-algo"               |
+| `Allocation` (string)                     | An Eppo allocation key                                                                                                   | "allocation-17"                     |
+
+
+## Usage with Contextual Multi-Armed Bandits
```
aarsilv, line 164: 💪

docs/sdks/server-sdks/dotnet.md
```diff
@@ -49,60 +75,262 @@ internal class AssignmentLogger : IAssignmentLogger
...
+    public void LogBanditAction(BanditLogEvent banditLogEvent)
+    {
+        analytics.Track("Eppo Bandit Action", banditLogEvent);
+    }
+}
 ```
-eppoClient.GetBoolAssignment(string subjectKey, string flagKey, SubjectAttributes? subjectAttributes = null)
-eppoClient.GetNumericAssignment(string subjectKey, string flagKey, SubjectAttributes? subjectAttributes = null)
+
+The SDK will invoke the `logBanditAction()` function with a `BanditLogEvent` object that contains the following fields:
```
aarsilv, line 202: I believe we want `LogBanditAction()` to be capitalized (silly dot net conventions 🙄 )

docs/sdks/server-sdks/dotnet.md
```diff
@@ -49,60 +75,262 @@ internal class AssignmentLogger : IAssignmentLogger
...
+     "flag-with-shoe-bandit",
+     "user123",
+     subjectAttributes,
+     actions,
+     "default");
+
+if (result.Action != null)
+{
+    // Follow the Bandit action
+    renderShoeAd(result.Action);
```
aarsilv, line 260: would dot net convention capitalize these method names?

docs/sdks/server-sdks/dotnet.md
```diff
@@ -49,60 +75,262 @@ internal class AssignmentLogger : IAssignmentLogger
...
+| `ActionProbability` (number)                         | The weight between 0 and 1 the bandit valued the assigned action                                                  | 0.25                             |
+| `OptimalityGap` (number)                             | The difference between the score of the selected action and the highest-scored action                             | 456                              |
+| `ModelVersion` (string)                              | Unique identifier for the version (iteration) of the bandit parameters used to determine the action probability   | "v123"                           |
+| `MetaData` IDict<string, string>                     | Any additional freeform meta data, such as the version of the SDK                                                 | `{ "sdkLibVersion": "3.5.1" }`   |
+
+
+### Querying for a Bandit Action
+To query the bandit for an action, there are multiple overloads of the `getBanditAction()` function you can use.  This function takes the following parameters:
```
aarsilv, line 223: You allude to overloads but then don't say anything else. I wonder if we introduce this method and then at the end of this subsection say we have some convenience overloads or something like that and briefly explain them.

docs/sdks/server-sdks/dotnet.md
```diff
@@ -49,60 +75,262 @@ internal class AssignmentLogger : IAssignmentLogger
...
+        // Need to reformat user attributes a bit; EppoClient requires `IDictionary<string, object?>`
+        var subjectAttributes = subjectTraits.Keys.ToDictionary(key => key, key => (object)subjectTraits[key]);
+        // Get an assignment for the user
+        var assignedVariation = eppoClient.GetStringAssignment(
+            "new-user-onboarding",
+            userID,
+            subjectAttributes,
+            "control"
+        );
+    }
```
aarsilv, line 314: what about also getting a bandit assignment after this? (if so since that is a string typed method, consider changing the above to something else like a boolean)

docs/sdks/server-sdks/dotnet.md
```diff
@@ -49,60 +75,262 @@ internal class AssignmentLogger : IAssignmentLogger
...
+    // User was not selected for a Bandit.
+    // A variation is still assigned.
+    renderDefaultShoeAd(result.Variation);
+}
 ```
 
-### Handling the null assignment
 
-We recommend always handling the empty assignment case in your code. Here are some examples illustrating when the SDK returns `null`:
+## Full Initialization and Assignment Example
```
aarsilv, line 269: Love this!

docs/sdks/server-sdks/dotnet.md
```diff
@@ -49,60 +75,262 @@ internal class AssignmentLogger : IAssignmentLogger
...
+    // Follow the Bandit action
+    renderShoeAd(result.Action);
+} else {
+    // User was not selected for a Bandit.
+    // A variation is still assigned.
+    renderDefaultShoeAd(result.Variation);
+}
 ```
 
-### Handling the null assignment
```
aarsilv, line 92: 🪓

docs/sdks/server-sdks/dotnet.md
```diff
@@ -3,39 +3,65 @@ import TabItem from '@theme/TabItem';
...
 var eppoClient = EppoClient.Init(eppoClientConfig);
 ```
 
 After initialization, the SDK begins polling Eppo's API at regular intervals to retrieve the most recent experiment configurations such as variation values and traffic allocation. The SDK stores these configurations in memory so that assignments thereafter are effectively instant. For more information, see the [architecture overview](/sdks/architecture/overview) page.
 
 If you are using the SDK for experiment assignments, make sure to pass in an assignment logging callback (see [section](#define-an-assignment-logger-experiment-assignment-only) below).
 
-### Define an assignment logger (experiment assignment only)
+#### Assign anywhere
+
```
aarsilv, line 40: I think we should have a table of the parameters like we do for Querying for a Bandit Action. For example, it is not clear what that fourth "control" argument is/does.

docs/sdks/server-sdks/dotnet.md
```diff
@@ -49,60 +75,262 @@ internal class AssignmentLogger : IAssignmentLogger
...
+| `ModelVersion` (string)                              | Unique identifier for the version (iteration) of the bandit parameters used to determine the action probability   | "v123"                           |
+| `MetaData` IDict<string, string>                     | Any additional freeform meta data, such as the version of the SDK                                                 | `{ "sdkLibVersion": "3.5.1" }`   |
+
+
+### Querying for a Bandit Action
+To query the bandit for an action, there are multiple overloads of the `getBanditAction()` function you can use.  This function takes the following parameters:
+- `flagKey` (string): The key of the feature flag corresponding to the bandit
+- `subject` (ContextAttributes): The key of the subject or user assigned to the experiment variation along with the subject's Categorical and Numeric attributes
+- `actions` (IDictionary<string, ContextAttributes>): Map of actions (by name) to their Context (Categorical and Numeric)
+- `defaultValue` (string): The default *variation* to return if the flag is not successfully evaluated
```
aarsilv, line 227: or--the more common scenario--the flag is disabled.

`The default *variation* to return if the flag is disabled or not successfully evaluated`

aarsilv (approved):
Thanks for giving this so much love! Approving assuming you agree with and act on my comments 😛

docs/sdks/server-sdks/dotnet.md
typotter: In my haste, I wrote a little like little Winnie The Pooh, I suppose, where bandits are a Very Important Thing. oh bother

docs/sdks/server-sdks/dotnet.md
typotter: ✔️

docs/sdks/server-sdks/dotnet.md
typotter: yes. Some conventions seem to be rooted just in doing it differently

docs/sdks/server-sdks/dotnet.md
typotter: done

docs/sdks/server-sdks/dotnet.md
typotter: done

docs/sdks/server-sdks/dotnet.md
typotter: done

docs/sdks/server-sdks/dotnet.md
```diff
@@ -49,60 +83,339 @@ internal class AssignmentLogger : IAssignmentLogger
...
+            userID,
+            subjectAttributes,
+            "control"
+        );
+
+        // Get a bandit action
+        var actions = new List<String>(){"nike", "adidas"};
+        var banditResult = eppoClient.GetBanditAction(
+          "shoe-bandit",
+          ContextAttributes.FromDictionary(userID, subjectAttributes),
```
aarsilv, line 380: Any reason you're passing this instead of just `subjectAttributes`?

typotter: That's actually one overload that's not available (subjectKey, subject Attributes and List<String> actions).

PR #513, opened 2024-11-04
SDK tags page

emetelka:


docs/feature-flagging/concepts/sdk-tags.md
```diff
@@ -0,0 +1,36 @@
+# SDK tags
+
+Tags allow you to organize your flags in the Eppo UI, but SDK tags include an additional benefit.
+
+Eppo creates a configuration file for the client, but when your program is using hundreds of flags across multiple applications, there may be a concern that you're unnecessarily increasing the size of the configuration file. SDK tags allow you to indicate which flags you'd like to include in the configuration file, reducing each the configuration file size.
```
lukasgoetzweiss, line 5:
```suggestion
Eppo creates a configuration file for the client, but when your program is using hundreds of flags across multiple applications, there may be a concern that you're unnecessarily increasing the size of the configuration file. SDK tags allow you to indicate which flags you'd like to include in the configuration file, reducing the configuration file size.
```

lukasgoetzweiss (approved):
Nice! At some point we'll want to make this discoverable from the SDK section too, probably in the "SDK keys and environments" pages. I can add that to the backlog

PR #510, opened 2024-10-30
feat: keys for anonymous subjects

sameerank:
https://deploy-preview-510--eppo-data-docs.netlify.app/sdks/sdk-features/subjects/

PR #515, opened 2024-11-05
Lg/ff 3436

lukasgoetzweiss:
Adapted the PR [here](https://github.com/Eppo-exp/eppo-docs/pull/510) to include more information about subjects and anonymous users, and breakout the practical solutions to anonymous IDs as a guide.

Considered adding rolling your own ID resolution that Heather wrote about [here](https://www.notion.so/eppo/Identity-Resolution-for-Feature-Flags-12349cc0114380f8b71ad1542ae963db?utm_content=12349cc0-1143-80f8-b71a-d1542ae963db&utm_campaign=T01U53NQR96&pvs=6), but I'd prefer to have a working example before putting into the docs. If we get a working example, it would be a natural addition to the guide proposed in this PR.

I also renamed the "SDK Features" section "Concepts" and "subjects" doesn't feel like a feature as much as it feels like a concept

### DIFF
--- a/docs/sdks/sdk-features/_category_.json
+++ b/docs/sdks/sdk-features/_category_.json
@@ -1,5 +1,5 @@
 {
-    "label": "SDK features",
-    "position": 7
+    "label": "Concepts",
+    "position": 3
   }
   
--- a/docs/data-management/certified-metrics/dbt-models.md
+++ b/docs/data-management/certified-metrics/dbt-models.md
@@ -123,7 +123,7 @@ jobs:
     - name: Install dependencies
       run: |
         python3 -m pip install --upgrade pip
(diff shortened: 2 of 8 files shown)


docs/sdks/sdk-features/_category_.json
```diff
@@ -1,5 +1,5 @@
 {
-    "label": "SDK features",
-    "position": 7
+    "label": "Concepts",
```
sameerank, line 2: Yup, agreed this is a better name

sameerank (approved):
This is great! Thank you for carrying this forward

hhargreaveseppo (approved):
lgtm! I added the vendor examples in as well

PR #524, opened 2024-11-13
Include mention of the Delta method in the Clustered Analysis page

bertilhatt:


tbuffington7 (commented):
We also have a page that explains the methodology in more detail [here](https://docs.geteppo.com/guides/advanced-experimentation/analyzing-clustered-experiments/)


Maybe we can move that page I linked to the Statistics section and then link to it in this page?

PR #525, opened 2024-11-14
Adjusting table of contents

nikomancy:
## What’s changing?
I’m making some adjustments to the TOC. Specifically, I’m making guides and SDKs into top-level categories in the TOC.

## Why?
The goal is to make the parts of our docs that speak to engineers more visible. In many cases, we do have docs for engineering needs; they are just hidden in collapsible sections. Bringing these to the surface will make it easier for developers and engineers to see these docs exist and to navigate to relevant information. This should improve developer experience and increase developer confidence in the docs.

## Testing
Tested on local.


lukasgoetzweiss (approved):
I like it!

PR #527, opened 2024-11-18
Adding init flag availability info to SDKs

nikomancy:
Per gap identified by @chasdevs.

Adding explicit information around when flags become available relative to initialization.

### DIFF
--- a/docs/sdks/client-sdks/android.md
+++ b/docs/sdks/client-sdks/android.md
@@ -34,6 +34,8 @@ CompletableFuture<EppoClient> eppoClientFuture = new EppoClient.Builder("YOUR_SD
     .buildAndInitAsync();
 ``` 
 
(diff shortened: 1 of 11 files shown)


docs/sdks/client-sdks/android.md
```diff
@@ -34,6 +34,8 @@ CompletableFuture<EppoClient> eppoClientFuture = new EppoClient.Builder("YOUR_SD
     .buildAndInitAsync();
 ``` 
 
+Flags and other features become available to the SDK after initialization.
```
chasdevs, line 37: I'd just rephrase it to "The SDK has access to flag configurations once initialization is complete."

chasdevs (approved):
I think this works. I'd wait for approval from an engineer on the feature flagging team to make sure it's correct for all SDKs.

PR #537, opened 2024-11-25
Adding API Option Reference Component

nikomancy:
## What
Adding a component to template API reference tables.

image

## Why

Markdown tables leave a lot to be desired for maintainability and authoring. Additionally, the HMTL tables they output have display issues on diminished viewports and also have some issues with inconsistent column sizing.

This component fixes both of those issues. First, the syntax doesn't require putting entire rows into single lines. Second, component lets you pass markdown and react components as part of its inner element for the description. Lastly, it templates out into a responsively designed set of divs.

```markdown
<ApiOptionRef name="Markup Test" defaultValue="5000">
Markdown syntax test

Lorem Ipsum

*italics*

**bold**

[link](https://www.google.com)


Table:

| Header 1 | Header 2 | Header 3 |
| -------- | -------- | -------- |
| Cell 1   | Cell 2   | Cell 3   |

\`\`\`js
// code block
\`\`\`

:::tip
Even other mdx elements work
:::

</ApiOptionRef>


```

Outputs: 

image

Diminished Viewport:

image

### DIFF
--- a/src/components/ApiOptionRef.js
+++ b/src/components/ApiOptionRef.js
@@ -1,0 +1,23 @@
+import React from 'react';
+import MDXContent from '@theme/MDXContent';
+
+function ApiOptionRef({ name, children, defaultValue }) {
+  return (
+    <div className="api-option">
+      <div className="api-option__header">
+        <code className="api-option__name">{name}</code>
+        {defaultValue && (
+          <span className="api-option__default">
+            Default: <code>{defaultValue}</code>
+          </span>
+        )}
+      </div>
+      
+      <div className="api-option__description">
+        <MDXContent>{children}</MDXContent>
+      </div>
+    </div>
+  );
+}
+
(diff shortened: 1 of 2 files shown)


nikomancy:
Made default field optional and added optional field for type that supports hyperlinking. There might be other variants of this component I need to add later depending on the tables it is replacing.

image

src/components/ApiOptionRef.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import MDXContent from '@theme/MDXContent';
+
+<<<<<<< Updated upstream
```
lukasgoetzweiss, line 4: I like the direction!

Looks like some conflicts might still need to be resolved.

nikomancy: Should be solved now.

PR #543, opened 2024-12-05
Niko/upgrade docusaurus2 to docusaurus3

nikomancy:
Updating Docusaurus through a major version to 3.6.4

This will open a number of new features to us. The big one is proper functioning on content partials which prior to 3.x did not properly work with the on page TOC.

### What's changed
- Everything runs on Docusaurus version 3 now.
- MDX now runs with stricter settings to avoid issues with React. The short version is that {, and <, will now need to be escaped by a backslash \< \{ or wrapped in code back ticks. `<` `{`. This came up a bit through the docs in our types and in a few math equations.


emetelka (approved):
Spot-checked on key pages and looks good

PR #558, opened 2024-12-19
Adding Python docs to new template update

nikomancy:
## What?

Updating the Python docs to follow the same multipage pattern as our Javascript docs.

## Why?

- The goal is to make the learning journey for engineers working with specific SDKs easier.
- Content is easier to find in the TOC, improving navigability.
- Code samples for typed assignment examples have been added for ease of use.
- Boilerplate prose docs have been added to allow engineers to learn about relevant conceptual information alongside the code without having to context switch to other parts of the docs.

## Review instructions

- Let's discuss the Python SDK 4.0 announcement doc. Are there any places in the docs where the differences between 3.0 and 4.0 are different enough that we need to preserve older doc information? Are we safe to get rid of that page?
- Please make sure the code samples are accurate.

## Todo before merge

- [ ] Delete python.md
- [ ] Delete announcing-python-4-0.md
- [ ] Update redirects

### DIFF
--- a/docs/sdks/server-sdks/python/examples.mdx
+++ b/docs/sdks/server-sdks/python/examples.mdx
@@ -1,0 +1,152 @@
+---
+title: Usage Examples
+sidebar_position: 6
+---
+
+In this section, we will go through a few examples of how to use Eppo's Python SDK in two common situations: batch processing and real-time applications.
+
+## Usage in a batch process
+
+As a basic example, imagine using Eppo's SDK to randomize users in a batch machine learning evaluation script. Further, imagine we're testing an upgrade from model version `v1.0.0` to `v1.1.0`. First we'd create a feature flag called `ml-model-version` with two variations: `v1.0.0` and `v1.1.0`. The Eppo UI should look like something like this:
+
+![Feature flag configuration](/img/feature-flagging/script_example_flag_config.png)
+
+:::note
+If you have not set up an experiment in Eppo's UI before, please see the [experiment quickstart guide](/experiment-allocation-quickstart).
+:::
+
+To keep the example simple, let's not worry about logging assignments to a warehouse and instead just write them to a local file using Python's built-in `logging` package.
+
+Your ML model evaluation script might then look something like this:
+
+```python title=script_example.py
+import logging
+import os
+from uuid import uuid4
+import eppo_client
+from eppo_client.config import Config, AssignmentLogger
+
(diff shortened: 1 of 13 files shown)


lukasgoetzweiss:
It looks like this PR drops some of [the examples](https://docs.geteppo.com/sdks/server-sdks/python/#examples) that are currently live. Can we add a page including those? This is a great example of where having a whole python section can help us add dedicated pages for common use cases: batch process jobs, django apps, etc.

Also, can we update this PR to include [this feedback](https://eppo-group.slack.com/archives/C058QMZ6XQC/p1734473810166919) from Slack?

rasendubi (commented):
Review: https://github.com/Eppo-exp/eppo-docs/pull/561

One of the diffs between pre/post 4.0 is that 4.0 exposes all things from `eppo_client` directly, whereas pre-4.0 exported things from somewhat random places. (Old locations are still supported but depricated.)

docs/sdks/server-sdks/python/examples.mdx
```diff
@@ -0,0 +1,152 @@
...
+
+    print(f"{user_id}: {model_version}")
+```
+
+Note that since the `get_string_assignment` call is deterministic, the same `user_id` will always produce the same `variation`. That is, if this script runs at a regular cadence, you don't need to worry about users switching between variations.
+
+After running this, you can inspect the logs in the `eppo_assignments.csv` file:
+
+```json title=eppo_assignments.csv
+{'base': {'featureFlag': 'ml-model-version', 'allocation': 'allocation-10061', 'experiment': 'ml-model-version-allocation-10061', 'variation': 'v1.1.0', 'metaData': {'sdkName': 'python', 'sdkVersion': '4.0.1', 'coreVersion': '4.0.0'}}, 'subject': 'f54ba5f3-90bb-4cdf-bf56-ca0335ede92c', 'subjectAttributes': {}, 'timestamp': '2024-10-11T03:24:12.093538Z'}
```
rasendubi, line 72: This output seems to be buggy because (1) it's not JSON line (json uses `"` instead of `'`), and (2) this has `base` as a separate field, which was a bug fixed in python-sdk 4.0.3 back in October

lukasgoetzweiss: https://github.com/Eppo-exp/eppo-docs/pull/574

PR #561, opened 2025-01-06
Review python docs

rasendubi:
This is a review of #558.


leoromanovsky (approved):
Thanks for structuring your feedback as a PR; nice to review it like this.

nikomancy (approved):
This looks good and helps a lot. I see the need to add missing fields to the table in assignments and I will tackle that before merging this to main.

PR #570, opened 2025-01-21
added assignment logger as an input for ClientConfig

lukasgoetzweiss:
The SDK will throw an error if `assignment_logger` is not specified in `ClientConfig`. This updates the quickstart and initialization pages to account for this.

I also made the imports more explicit for the experiment and bandit sections of the quickstart, in case the reader skipped over the flagging section


tbuffington7 (approved):
LGTM. Thanks for updating

PR #584, opened 2025-01-31
Added note about "use client" to React docs

lukasgoetzweiss:
We have dedicated Next.js docs, but I figure this is easy enough to get tripped up on that it's worth adding a callout

### DIFF
--- a/docs/sdks/client-sdks/javascript/react.mdx
+++ b/docs/sdks/client-sdks/javascript/react.mdx
@@ -43,7 +43,7 @@ export default function EppoRandomizationProvider({
 ```
 
 :::note
-If you are using Next.js, make sure this component is rendered client side by adding `"use client"` to the top of the file. For more details on using Eppo in Next.js, see (here)[/sdks/client-sdks/javascript/nextjs-setup/].
+If you are using Next.js, make sure this component is rendered client side by adding `"use client"` to the top of the file. For more details on using Eppo in Next.js, see [here](/sdks/client-sdks/javascript/nextjs-setup/).
 :::
 
 After the SDK is initialized, you may assign variations from any child component of `EppoRandomizationProvider`. We recommend wrapping the SDK code in a [useMemo hook](https://reactjs.org/docs/hooks-reference.html#usememo) to avoid invoking the assignment logic on every render:
--- a/docs/sdks/client-sdks/javascript/react.mdx
+++ b/docs/sdks/client-sdks/javascript/react.mdx
@@ -42,6 +42,10 @@ export default function EppoRandomizationProvider({
 }
 ```
 
+:::note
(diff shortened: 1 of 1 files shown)


aarsilv (commented):
What sparked this change--I don't think Next.js would be at play if React Native is. Did you mean to make a modification in our javascript client docs? If so, perhaps you could also include a "see this next.js guide for more detail"

lukasgoetzweiss:
> What sparked this change--I don't think Next.js would be at play if React Native is. Did you mean to make a modification in our javascript client docs? If so, perhaps you could also include a "see this next.js guide for more detail"

This is in the javascript docs, just on the dedicated React page (not React Native)

The prompt for this was that I was setting up a Next JS app and going off of the React docs (I ended up here from the link "using with React" in the flagging quick start).

If I was really paying attention, I would have seen the "Next JS" thing in the sidebar, but from my (very limited) understanding of the JS universe, it seems like it's fairly possible to imagine another reader going through this same arc.

Lmk if you disagree though

I can add a link to the Next JS docs too, that's a good call

docs/sdks/client-sdks/javascript/react.mdx
```diff
@@ -42,6 +42,10 @@ export default function EppoRandomizationProvider({
 }
 ```
 
+:::note
+If you are using Next.js, make sure this component is rendered client side by adding `"use client"` to the top of the file.
```
lukasgoetzweiss, line 46:
```suggestion
If you are using Next.js, make sure this component is rendered client side by adding `"use client"` to the top of the file. For more details on using Eppo in Next.js, see (here)[/sdks/client-sdks/javascript/nextjs-setup/].
```

docs/sdks/client-sdks/javascript/react.mdx
```diff
@@ -42,6 +42,10 @@ export default function EppoRandomizationProvider({
 }
 ```
 
+:::note
+If you are using Next.js, make sure this component is rendered client side by adding `"use client"` to the top of the file. For more details on using Eppo in Next.js, see (here)[/sdks/client-sdks/javascript/nextjs-setup/].
```
typotter, line 46:
```suggestion
If you are using Next.js, make sure this component is rendered client side by adding `"use client"` to the top of the file. For more details on using Eppo in Next.js, see [here](/sdks/client-sdks/javascript/nextjs-setup/).
```

typotter (commented):
Thanks for the update.
Just needs a small change to fix the link in your note.

typotter (approved):

aarsilv (approved):
> This is in the javascript docs, just on the dedicated React page (not React Native)

Oh right! Expectation bias. This is great, thanks for doing this!

PR #609, opened 2025-02-26
analysis only braze docs

hhargreaveseppo:


docs/guides/marketing/braze.md
```diff
@@ -0,0 +1,46 @@
...
+- Selecting the message type (e.g., Push, Email, In-App Message, etc.).
+- Defining the target audience for your A/B test.
+- Configuring the variations.
+- Scheduling the campaign.
+
+Once your experiment is set up, launch the test in Braze.
+
+### **Eppo Experiment setup**
+
+The first step to setting up your experiment in Braze is to create an [Assignment table](/data-management/definitions/assignment-sql) in Eppo for the user level Braze campaign data in your warehouse. Use [Braze’s docs](https://www.braze.com/docs/user_guide/data/braze_currents/event_glossary/message_engagement_events?tab=cloud%20storage) to determine which event schema you should use and align with the data requirements with your Eppo Assignments table.
```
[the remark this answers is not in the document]
lukasgoetzweiss: I found this paragraph confusing. Shouldn't it be a consistent table the people will have to pull from? 

"Experiment split entry events" looks the most promising based on the Braze docs you linked to

docs/guides/marketing/braze.md
```diff
@@ -0,0 +1,46 @@
...
+- Configuring the variations.
+- Scheduling the campaign.
+
+Once your experiment is set up, launch the test in Braze.
+
+### **Eppo Experiment setup**
+
+The first step to setting up your experiment in Braze is to create an [Assignment table](/data-management/definitions/assignment-sql) in Eppo for the user level Braze campaign data in your warehouse. Use [Braze’s docs](https://www.braze.com/docs/user_guide/data/braze_currents/event_glossary/message_engagement_events?tab=cloud%20storage) to determine which event schema you should use and align with the data requirements with your Eppo Assignments table. 
+
+When analyzing channel outreach experiments, it's common to measure both overall user level metrics (engagement, revenue, retention, etc.) and campaign-specific metrics (click through rate, unsubscribes, etc.). Depending on the entities that your metrics use, you may want to include a [Secondary Entity](/data-management/definitions/assignment-sql/#optional-columns-for-advanced-use-cases) to include metrics that use a different user id than what is used in Braze.
```
lukasgoetzweiss, line 32:
```suggestion
When analyzing channel outreach experiments, it's common to measure both overall user-level metrics (engagement, revenue, retention, etc.) and campaign-specific metrics (click through rate, unsubscribes, etc.). Depending on the entities that your metrics use, you may want to include a [Secondary Entity](/data-management/definitions/assignment-sql/#optional-columns-for-advanced-use-cases) to include metrics that use a different user id than what is used in Braze.
```

docs/guides/marketing/braze.md
```diff
@@ -0,0 +1,46 @@
...
+
+| Column | Type in Eppo | Data from Braze |
+| --- | --- | --- |
+| **`assignment_timestamp`** | Timestamp | **`time`** |
+| **`user_id`** | Experiment subject ID | **`user_id`** Braze defined user id, this may also be **`external_user_id`** if you are managing user ids in Braze |
+| **`experiment`** | Experiment key | **`campaign_id`** |
+| **`variant`** | Variant | **`message_variation_id`** Might be slightly different based on the Braze event schema that you use. |
```
lukasgoetzweiss, line 39: Where does this variant field live? I see `canvas_variation_name` on the experiment splits entry events table, but not sure if I should be looking somewhere else

docs/guides/marketing/braze.md
```diff
@@ -0,0 +1,46 @@
...
+| **`email_send_id` (optional, use if needed)** | Secondary ID (opt to use a Secondary Entity if you want to include additional metrics that are bridged to a different ID not utilized in your Braze Schema) | Universal internal id from bridged from an internal table: **`internal_user_id`**  |
```
lukasgoetzweiss, line 40: Similarly, what event is this from? i see `send_id` on "Email send events". Is that what this is referencing?

docs/guides/marketing/braze.md
```diff
@@ -0,0 +1,46 @@
...
+
+Experiment analyses that use this assignment source can measure both user-level metrics and campaign specific metrics.
+
+After the Braze Assignment table is set up, set up any additional [Facts](/data-management/definitions/fact-sql/) and [Metrics](/data-management/metrics/) that should be included with Braze experiments analyzed in Eppo including clicks or responses.
```
lukasgoetzweiss, line 44: We should consider adding one example here too

lukasgoetzweiss (commented):
Left a few comments. I think more specific recommendations to their docs would make this read better. 

I saw Braze has a free trial. Wonder if we should spin that up and get this example working end to end

docs/guides/marketing/braze.md
hhargreaveseppo: Really good call out + question.

My understanding is that a Canvas in Braze equates to a [multi-step campaign](https://www.braze.com/docs/user_guide/engagement_tools/canvas/create_a_canvas/create_a_canvas/#editing-connections) that can be created in the UI. Users can have multiple legs of qualification, ie user is in segment 1, gets a message, does another event that qualifies them for segment 2 and now is eligible for receiving the next step in the journey only after having completed the first step. Segment has something similar called Journeys. 

A Campaign in Braze is a single time entry. Ie I want to target users in this cohort and they can receive this series of messages.

I don't have enough feedback from customers to know if they're typically using a Canvas or a Campaign to create their test. In Braze there's a different workflow for creating a test for a [Canvas](https://www.braze.com/docs/user_guide/engagement_tools/canvas/testing_canvases/sending_test_canvases) vs a [Campaign](https://www.braze.com/docs/user_guide/engagement_tools/testing/multivariant_testing). I'm assuming the canvas_variation_id could be null on a an email delivery event if that email was sent through a campaign and not a canvas.

Textnow's workspace is a good example of how to manage both the Canvas and Campaign tests -- they've created a separate assignment table for both and use the corresponding canvas_variation_id or message_variation_id depending on how the test was created.

I can include a distinction on canvas vs campaign tests

docs/guides/marketing/braze.md
hhargreaveseppo: ack yea i got confused here. I was thinking that `send_id` equated to a uuid for an id for the email send; I'll update this to reflect the combined user_id + email_campaign_id identifier we discussed last week.

docs/guides/marketing/braze.md
```diff
@@ -0,0 +1,111 @@
...
+
+
+### **Including Campaign and Canvas Specific Metrics with a Secondary Entity**
+
+When analyzing channel outreach experiments, it's common to measure both overall user-level metrics (engagement, revenue, retention, etc.) and campaign-specific metrics (click through rate, unsubscribes, etc.). 
+
+A [Secondary Entity](/data-management/definitions/assignment-sql/#optional-columns-for-advanced-use-cases) on the Assignments table will ensure any message-specific metric such as Clicks, Bounces, or Opens are specific to the Campaign or Canvas experiment.
+
+To include Campaign or Canvas specific metrics:
+1. [Create a new Entity](/data-management/definitions/entities/#creating-an-entity) called **Combined ID**.
```
lukasgoetzweiss, line 88: Would "User-Campaign ID" be a more clear name?

docs/guides/marketing/braze.md
```diff
@@ -0,0 +1,111 @@
...
+2. On the Braze Campaign or Canvas Assignments table, concatenate the User ID and Campaign or Canvas Name and adding it as a Secondary Entity called `combined_id` For example, a Campaign called **Email_14_day_nurture**, would be concatenated to `12345-Email_14_day_nurture` as the `combined_id`.
```
lukasgoetzweiss, line 89:
```suggestion
2. On the Braze Campaign or Canvas Assignments table, concatenate the User ID and Campaign or Canvas Name and adding it as a Secondary Entity called `combined_id` For example, user `12345` enrolling into a campaign called **Email_14_day_nurture**, would be concatenated to `12345-Email_14_day_nurture` as the `user_campaign_id`.
```

lukasgoetzweiss: (just to make it a bit clearer where 12345 comes from)

lukasgoetzweiss (approved):
lgtm!

PR #617, opened 2025-03-06
added cuped stratitifcation section

lukasgoetzweiss:


docs/statistics/cuped.md
```diff
@@ -53,7 +53,17 @@ This is often the most important variable in a regression, but does suffer some
...
+
+## CUPED++ and stratification
+
+Stratification is a common technique used in experiment design. By ensuring that treatment and control groups have a similar distribution of categorical features (prior engagement data, demographics, etc.), we can run experiments with tighter confidence intervals.
+
+In practice however, online stratification systems tend to lead to engineering overhead and performance issues. Other options such as seed finding also have substantial limitations, namely that they extend poorly to new users. 
+
+Fortunately, Eppo's CUPED++ model can give mathematically equivalent estimates as stratification, but without the engineering or performance overhead. This is referred to as **post hoc stratification**. Simply add the categorical variables on which you want to balance as assignment properties and you will get a variance reduction equivalent to that from a stratified design. Not only does this alleviate performance issues, it also allows you to include as many variables as you'd like, even if they are not available at the time of assignment.
```
schmit, line 64:
```suggestion
Fortunately, Eppo's CUPED++ model can give mathematically equivalent estimates as stratification, but without the engineering or performance overhead. This is referred to as **post-hoc stratification**. Simply add the categorical variables on which you want to balance as assignment properties and you will get a variance reduction equivalent to that from a stratified design. Not only does this alleviate performance issues, it also allows you to include as many variables as you'd like, even if they are not available at the time of assignment.
```

schmit (approved):
🚢 🇮🇹

PR #628, opened 2025-03-13
Add switchback docs

emetelka:
I still need to add images


docs/experiment-analysis/switchbacks.md
```diff
@@ -0,0 +1,78 @@
+# Switchback experiments
+
+:::note
+Switchback experiments are currently in closed Beta
+:::
+
+Marketplaces present unique challenges when it comes to experimentation, especially because they depend on complex interactions between multiple groups. Switchback tests solve this problem by randomizing entire environments (e.g., regions and time periods) instead of focusing on groups of users. For example, in a ride-sharing app, you might alternate between treatment (lower prices for riders) and control (regular prices) during a specific time of day across different cities. This way, all drivers and riders experience the same conditions for each phase, minimizing interference and isolating the effects of price changes cleanly. Switchback tests can simulate real-world conditions without fracturing the ecosystem or leaving groups with noticeably worse experiences.
```
DemetriPananoss, line 7:
```suggestion
Marketplaces present unique challenges when it comes to experimentation, especially because they depend on complex interactions between multiple groups. Switchback tests solve this problem by randomizing entire environments (e.g., regions and time periods) instead of focusing on groups of users. For example, in a ride-sharing app, you might alternate between pricing algorithms (new algorithm as treatment, old algorithm as control) during a specific time of day across different cities. This way, all drivers and riders experience the same conditions for each phase, minimizing interference and isolating the effects of price changes cleanly. Switchback tests can simulate real-world conditions without fracturing the ecosystem or leaving groups with noticeably worse experiences.
```

DemetriPananoss: Small nit.

docs/experiment-analysis/switchbacks.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Creating a Switchback Assignment SQL
+
+We'll now provide a step-by-step walkthrough for creating Switchback Assignment SQL Definitions in Eppo's UI.
+1. Navigate to Definitions, click the "Create definition SQL" button, and select Switchback Assignment SQL.
+2. Select the subject of the Switchback Assignment. This is the unit the Switchback is randomizing over, such as market or region.
+3. Write SQL in the SQL editor that returns assignment data from your data warehouse and click **Run**. 
+4. After clicking **Run**, you'll see some sample data. Annotate these columns into Eppo's data model using the right panel:
+5. Once you've finished annotating columns, click **Save & Close**
+
```
DemetriPananoss, line 36: We should probably explain here that the fact SQL needs to be at the subentity level and not the entity (e.g. market or region) level.

@bryanZavestoski we fell for this a couple of times.

lisaah (approved):

docs/experiment-analysis/switchbacks.md
```diff
@@ -0,0 +1,78 @@
...
+### Decision criteria
+
+The decision criteria sets the key metrics to be measured in the experiment and aligns the team around a ship decision based on the measurement of those metrics. See [Decision Criteria](/experiment-analysis/configuration/protocols/#decision-criteria/) for more information on configuring this part of the analysis.
+
+### Key, Variations, & Analysis dates
+
+This is where the experiment key is chosen, the variations are defined, and the dates for the experiment are configured.
+
+Of note for Switchbacks only, you can optionally add burn in and burn out periods. A burn period refers to a designated time interval at the start of a testing cycle where data is intentionally disregarded. This period allows the system to stabilize after switching between test conditions, such as transitioning from control to treatment in a switchback test. For example, when alternating pricing structures in a ride-sharing app, a burn period gives the marketplace time to adjust rider demand and driver availability before meaningful data is collected.
```
DemetriPananoss, line 64:
```suggestion
Of note for Switchbacks only, you can optionally add burn in and burn out periods. A burn period refers to a designated time interval at the start of a testing cycle where  assignment data is intentionally disregarded. This period allows the system to stabilize after switching between test conditions, such as transitioning from control to treatment in a switchback test. For example, when alternating pricing algorithms in a ride-sharing app, a burn period gives the marketplace time to adjust rider demand and driver availability before meaningful data is collected.
```

DemetriPananoss: Add small detail for what burn out is.

docs/experiment-analysis/switchbacks.md
emetelka:
```suggestion
### Updates to FactSQL
FactSQL that corresponds to metrics you want to monitor also need to be updated with the Switchback entity. On the FactSQL page, click edit and add a new subentity that corresponds to the Switchback entity. 
```

docs/experiment-analysis/switchbacks.md
emetelka: Question I got: If a region does not switch it's variation between time periods, does it still have a burn in and burn out period?

PR #632, opened 2025-03-14
 Fresh flag configs guide

emetelka:


docs/guides/engineering/fresh-config.md
```diff
@@ -0,0 +1,78 @@
...
+## Client-Side Applications
+
+### JavaScript Client
+For browser-based applications, you have several options to ensure fresh configurations:
+
+1. **Enable Polling**
+```javascript
+const client = new EppoClient({
+    apiKey: 'your-api-key',
+    pollInterval: 30000 // Poll every 30 seconds
```
aarsilv, line 18: This will set the polling interval but won't have it poll. replace with
`pollAfterSuccessfulInitialization: true`

(source: https://github.com/Eppo-exp/js-sdk-common/blob/main/src/client/eppo-client.ts#L87)

docs/guides/engineering/fresh-config.md
```diff
@@ -0,0 +1,78 @@
...
+});
+```
+
+2. **Manual Refresh**
+```javascript
+// Force a configuration refresh when needed
+await eppoClient.load();
```
aarsilv, line 25: for JavaScript you re-initialize with `forceReinitialize: true` argument in addition to your others.

You can see how we do it in a test file here: https://github.com/Eppo-exp/js-client-sdk/blob/main/src/index.spec.ts#L204

docs/guides/engineering/fresh-config.md
```diff
@@ -0,0 +1,78 @@
...
+### Mobile Applications
+Mobile apps can implement similar strategies:
+
+1. **Configure Polling**
+```swift
+// iOS Example
+let client = EppoClient(
+    apiKey: "your-api-key",
+    pollInterval: 30 // seconds
+)
```
aarsilv, line 37: iOS can't poll but you can fore a refresh with `resetSharedInstance()` and then call initialize again

Example in test: https://github.com/Eppo-exp/eppo-ios-sdk/blob/c1d2ad093d730dd7ddb2f2b6b5be5bffba5b650e/Tests/eppo/SingletonTest.swift#L16

docs/guides/engineering/fresh-config.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+```kotlin
+// Android Example
+val client = EppoClient.builder()
+    .apiKey("your-api-key")
+    .pollInterval(30) // seconds
+    .build()
```
aarsilv, line 45: for android and kotlin you'd use `.forceReinitialize(true)` to do a pushed reset

method is here: https://github.com/Eppo-exp/android-sdk/blob/main/eppo/src/main/java/cloud/eppo/android/EppoClient.java#L195

aarsilv, line 45: for android and kotlin you'd use `.forceReinitialize(true)` to do a pushed reset

method is here: https://github.com/Eppo-exp/android-sdk/blob/main/eppo/src/main/java/cloud/eppo/android/EppoClient.java#L195

docs/guides/engineering/fresh-config.md
```diff
@@ -0,0 +1,78 @@
...
+  - Implement graceful transitions when flags change
+
+### Real-Time Updates
+For the cases when real-time updates are necessary, we recommend adding a listener for a push from the server to force a refresh:
+
+1. **WebSocket Integration**
+```javascript
+// Example of listening to a WebSocket for configuration updates
+webSocket.on('config-update', async () => {
+    await client.refreshConfiguration();
```
aarsilv, line 62: I do like this interface though! We should consider something like this as syntatic sugar for reinitializing with same config and force reinit 😅

docs/guides/engineering/fresh-config.md
```diff
@@ -0,0 +1,81 @@
...
+)
+```
+
+```kotlin
+// Android Example
+val client = EppoClient.builder()
+    .forceReinitialize(true)
+    .apiKey("your-api-key")
+    .pollInterval(30) // seconds
+    .build()
```
typotter, line 48:
```suggestion
val client = new EppoClient.Builder("your-sdk-key", getApplication())
    .forceReinitialize(true)
    .pollingEnabled(true)
    .pollingIntervalMs(30 * 1000) // seconds
    .build()
```

typotter (commented):
Sorry for the fly-by review. Just noticed some issues with the code samples 

I would leave the **RealTime** updates out for now though as we don't have a way for either servers or clients to be notified in realtime when their configuration updates. We have a forthcoming SDK feature, `onConfigurationChange` that would allow a server to _observe_ when the config changes and then push to their client app via w/e push mechanism they have in place. That method is currently only available on java server (though node-server is in-progress)

aarsilv (approved):
Thanks for iterating! Approving assuming you factor in Tyler's suggestions.

PR #642, opened 2025-03-19
Clarify always full refresh for facts

emielver:


emetelka (approved):

docs/data-management/definitions/fact-sql.md
```diff
@@ -23,9 +23,10 @@ Each Fact SQL Definition should return the following columns. Note that the spec
...
 | Fact property (optional) | A categorical variable associated with the event. This can be used to filter events in metric creation, or to break out experiment results. Note that properties that are 1:1 with experiment subjects should be defined in an [Assignment](/data-management/definitions/assignment-sql) or [Entity Property](/data-management/definitions/property-sql) SQL definition. | `revenue_category` , `support_ticket_reason` , `event_name` |
 | Partition date (optional) | An optional additional timestamp for filtering rows with a column other than the event timestamp. Useful if your event timestamp column differs from the table's partition timestamp column. See [here](/data-management/warehouse-best-practices/#leveraging-partitioning) for more information about leveraging partitioning. | `date` |
+| Always full refresh | An optional boolean parameter that specifies if a metric associated to this source will fully refresh whenever the underlying SQL in this source changes. | `false` |
```
lukasgoetzweiss, line 26: This table is explaining the columns in your fact data model, I don't think it makes sense to put this option in this table

docs/data-management/definitions/fact-sql.md
```diff
@@ -168,12 +169,12 @@ You can update facts by clicking the `Edit` button to access the Fact SQL. At th
 Pressing the `Run` button will enable the mapping fields. Click `Save & Close` button to save any changes made in either the SQL or mapping.
 
 :::note
-Any running experiments with metrics based on the updated Fact SQL will automatically fully refresh those metrics on the next experiment update.
+Any running experiments with metrics based on the updated Fact SQL will automatically fully refresh those metrics on the next experiment update, if "Always full refresh" is enabled for that Fact SQL.
```
lukasgoetzweiss, line 172: On fact edit, we will still refresh experiments from scratch regardless of whether the "always full refresh" button is selected as long as they select it in the new modal that we've recently added.

Agreed this section should be updated, but it should describe the new modal we added (which lets users specify which metrics to refresh on next scheduled run)

lukasgoetzweiss (changes requested):

docs/data-management/definitions/fact-sql.md
lukasgoetzweiss: @emielver I think this comment was missed before merging. We always will give the option to fully refresh results on fact SQL update. It does not depend on whether "always full refresh" was enabled.

emielver: Ah yes, you're right. Great catch!

emielver: Updated it [here](https://github.com/Eppo-exp/eppo-docs/pull/673)

PR #651, opened 2025-03-28
Add Architecture and Security page to the Reference section

lukasgoetzweiss:


emetelka (approved):
Looks good. We will have to update the flag section with information about the event logger, but can tackle that in the future

PR #658, opened 2025-04-10
Webhook signature verification instructions

sameerank:
[Link to linear](https://linear.app/eppo/issue/FF-4298/webhook-signature-verification-instructions)

Clarifications/corrections:

* Unclear what "payload" means: the entire event-data-signature object? But how could we possibly generate a signature if the payload already has a signature in it? It would be less confusing if we specify to use the "data" object in the request body.
* JSON data types must not be pre-coerced into strings before running through the signature matching algorithm. This was a subtle gotcha that it would be preferable to avoid in the future.

### DIFF
--- a/docs/reference/webhook.mdx
+++ b/docs/reference/webhook.mdx
@@ -116,7 +116,7 @@ In short, any operation visible in the Flag History is sent to the webhook.
 
 ## Configuring the webhook
 
-To use the webhook, you first need to setup a URL on your side to receive the webhook payload.
+To use the webhook, you first need to setup a URL on your side to receive the webhook request.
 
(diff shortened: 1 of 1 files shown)


docs/reference/webhook.mdx
```diff
@@ -20,7 +20,7 @@ you can filter to just the events you are interested in subscribing to based on
 {
   "event": "experiment.metric.updated",
   "data": {
-    "experiment_id": "<experiment_id>",
+    "experiment_id": <experiment_id>,
```
sameerank, line 23: This is no longer valid JSON, but I wanted to make it clear to not expect a string representation of a numeric id.

aarsilv (approved):
🙌

PR #667, opened 2025-04-21
Polling in iOS

sameerank:
Related PR https://github.com/Eppo-exp/eppo-ios-sdk/pull/60

### DIFF
--- a/docs/sdks/client-sdks/ios/initialization.mdx
+++ b/docs/sdks/client-sdks/ios/initialization.mdx
@@ -79,6 +79,50 @@ Task {
 }
 ```
 
+#### Automatic Configuration Updates (Polling)
+
+The SDK can be configured to automatically poll for configuration updates at regular intervals. This is useful for ensuring your application always has the latest flag configurations without requiring manual updates.
+
+To enable polling during initialization:
+
+```swift
+Task {
+    try await EppoClient.initialize(
+        sdkKey: "SDK-KEY-FROM-DASHBOARD",
+        pollingEnabled: true,
+        pollingIntervalMs: 300_000,  // 5 minutes (default)
+        pollingJitterMs: 30_000      // 10% of polling interval (default)
+    )
+}
+```
+
+You can also start and stop polling after initialization:
+
+```swift
+let eppoClient = EppoClient.shared()
+
+// Start polling with custom intervals
+try await eppoClient.startPolling(
+    intervalMs: 300_000,  // 5 minutes
+    jitterMs: 30_000      // Jitter of up to 30 seconds
+)
+
+// Stop polling
+eppoClient.stopPolling()
+```
+
+The polling mechanism includes the following features:
+- Default polling interval of 5 minutes (300,000 ms)
+- Random jitter to prevent thundering herd problems (default is 10% of polling interval)
+- Exponential backoff on failures
+- Automatic retry up to 7 times on failed requests
+- Polling stops automatically if max retries are exceeded
+
+:::note
+For battery efficiency in mobile applications, it's recommended to stop polling when your app goes to the background and resume when it comes to the foreground.
+:::
+
+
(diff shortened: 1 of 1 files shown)


docs/sdks/client-sdks/ios/initialization.mdx
```diff
@@ -79,6 +79,49 @@ Task {
...
+
+// Stop polling
+eppoClient.stopPolling()
+```
+
+The polling mechanism includes the following features:
+- Default polling interval of 5 minutes (300,000 ms)
+- Random jitter to prevent thundering herd problems (default is 10% of polling interval)
+- Exponential backoff on failures
+- Automatic retry up to 7 times on failed requests
```
aarsilv, line 118: consider swapping the order of these two bullets

aarsilv (approved):

docs/sdks/client-sdks/ios/initialization.mdx
```diff
@@ -79,6 +79,49 @@ Task {
...
+- Polling stops automatically if max retries are exceeded
+
+:::note
+For battery efficiency in mobile applications, it's recommended to stop polling when your app goes to the background and resume when it comes to the foreground.
```
typotter, line 122: How about an example? Cursor gave me a number of suggestions

```swift
class AppDelegate: UIResponder, UIApplicationDelegate {
    
    var window: UIWindow?
    var eppoClient: EppoClient?
    
    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
        // Set up notification observers for app lifecycle events
        setupAppLifecycleObservers()
        
        // Initialize the Eppo client with polling enabled
        Task {
            do {
                eppoClient = try await EppoClient.initialize(
                    sdkKey: "YOUR_SDK_KEY",
                    pollingEnabled: true,
                    pollingIntervalMs: 300_000, // 5 minutes
                    pollingJitterMs: 30_000     // 30 seconds of jitter
                )
                
                print("Eppo client initialized with polling enabled")
            } catch {
                print("Failed to initialize Eppo client: \(error)")
            }
        }
        
        return true
    }
    
    private func setupAppLifecycleObservers() {
        // Register for app entering background
        NotificationCenter.default.addObserver(
            self,
            selector: #selector(appDidEnterBackground),
            name: UIApplication.didEnterBackgroundNotification,
            object: nil
        )
        
        // Register for app coming to foreground
        NotificationCenter.default.addObserver(
            self,
            selector: #selector(appWillEnterForeground),
            name: UIApplication.willEnterForegroundNotification,
            object: nil
        )
    }
    
    @objc private func appDidEnterBackground() {
        print("App entered background, stopping Eppo polling")
        
        Task {
            // Stop polling when app is in background to conserve resources
            await eppoClient?.stopPolling()
        }
    }
    
    @objc private func appWillEnterForeground() {
        print("App will enter foreground, restarting Eppo polling")
        
        Task {
            // Resume polling when app comes back to foreground
            await eppoClient?.startPolling()
        }
    }
    
    deinit {
        // Remove observers when app delegate is deallocated
        NotificationCenter.default.removeObserver(self)
    }
}
```

typotter (approved):
Thanks for rocking the docs!

docs/sdks/client-sdks/ios/initialization.mdx
leoromanovsky: Is that a swift or objc example?

typotter: Swift.

sameerank: I like the idea of a code example, but I think I need to understand this example a little better before using it, e.g. if the `@objc` decorator is necessary https://developer.apple.com/documentation/swift/migrating-your-objective-c-code-to-swift/#Migrate-Your-Code

PR #672, opened 2025-04-23
Add some more detail to flag allocations

emielver:


docs/feature-flagging/concepts/flag-allocations.md
```diff
@@ -15,7 +15,27 @@ In Eppo, allocations are stacked in a vertical waterfall as seen below:
...
+
+## Traffic exposure adjustments
+
+When you adjust the traffic exposure of an allocation:
+
+-   **Increasing traffic exposure**: When you increase traffic exposure, existing users who were already assigned to the allocation will maintain their variant assignments when they return (i.e. have sticky assignments). This means that users who have previously been bucketed for this allocation will continue to receive the same experience. Only new users (who are being evaluated for the first time) and users who were not previously bucketed for this allocation will be subject to the updated traffic rules.
+
+-   **Decreasing traffic exposure**: When you decrease traffic exposure, existing users who were already assigned to a bucket will be re-evaluated. This means that users who have previously been bucketed for a specific variant will potentially receive a different experience.
+
+-   **Experiment completion**: This is particularly useful for experiments with a measurement window. For example, if you've reached your sample size goals, you can set traffic exposure to 0% to stop new assignments while allowing already-assigned users to complete their measurement window.
```
lukasgoetzweiss, line 31: I think this section misrepresents how traffic exposure adjustments work. For instance, changing the traffic allocation to 0% will pass everyone through to the next assignment/allocation (the default variant in most cases), even if they had previously been enrolled.

Also, it's important to state that users will only receive the same variant if they remain in the targeted audience.

For example, imagine you have a flag where free users are randomly split between control and treatment, but paid users "fall through" to the default variant (control). In this case once a user goes from free -> paid, they will stop seeing the treatment. Eppo's SDK does not store any history of who has already been assigned.

Assignments are "idempotent", but not "sticky". That is, for the same inputs to the SDK (user ID, user context), the SDK will return the same value. But if a user's context changes and they are no longer in the targeted audience, Eppo's SDK will not "remember" that they were in treatment earlier.

The SDK does guarantee that if a user was in treatment when traffic exposure was at 5%, that they'll still be in treatment once we ramp to 20% exposure. This is because the hashing for traffic allocation and variant weighting is independent. I think that's what this section is trying to say, but we should be more clear about the limitations

lukasgoetzweiss (changes requested):

PR #692, opened 2025-07-22
docs: update iOS SDK docs for non-throwing assignment methods

sameerank:
Remove try keyword from `getStringAssignment`, `getBooleanAssignment`, `getIntegerAssignment`, `getNumericAssignment`, and `getJSONStringAssignment` method calls since these methods no longer throw exceptions.


aarsilv (approved):
📈

